PR #6, opened 2022-04-28
Create Dockerfile

didintri196:


aldinokemal (commented):
Hello, I don't think this is the best scenario for building a docker image, because it makes the image size very big. Could you change the golang images as builder then copy compiled app to linux scratch/alpine and install libvps instead of directly using golang images

aldinokemal:
https://github.com/aldinokemal/go-whatsapp-web-multidevice/pull/7

PR #60, opened 2023-02-08
feat: add reaction

aldinokemal:
### Context

- Add send reaction message


muava12:
how to delete reaction? if i set to blank it give me an error

PR #66, opened 2023-03-16
feat: Add Join Group with Link

tim-vandecasteele:
First of all, thanks for this great project, it was exactly what I was looking for!

I know very little of go. Almost nothing actually, but I wanted to add functionality to join a group with a link. This is exactly what I did in this pull request. 

I have no clue about testing or other pull request hygiene for this project, so let me know if you're interested in this change, and if you are, what should happen to make it ok for consumption.

### DIFF
--- a/src/internal/rest/group.go
+++ b/src/internal/rest/group.go
@@ -12,7 +12,7 @@ type Group struct {
 
 func InitRestGroup(app *fiber.App, service domainGroup.IGroupService) Group {
 	rest := Group{Service: service}
-	app.Post("/group/join_with_link", rest.JoinGroupWithLink)
+	app.Post("/group/join-with-link", rest.JoinGroupWithLink)
 	return rest
 }
 
--- a/src/internal/rest/group.go
+++ b/src/internal/rest/group.go
@@ -1,0 +1,33 @@
+package rest
+
+import (
+	domainGroup "github.com/aldinokemal/go-whatsapp-web-multidevice/domains/group"
+	"github.com/aldinokemal/go-whatsapp-web-multidevice/pkg/utils"
+	"github.com/gofiber/fiber/v2"
+)
+
+type Group struct {
+	Service domainGroup.IGroupService
+}
+
+func InitRestGroup(app *fiber.App, service domainGroup.IGroupService) Group {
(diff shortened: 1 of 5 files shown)


src/services/group.go
```diff
@@ -0,0 +1,32 @@
...
+
+
+func (service groupService) JoinGroupWithLink(_ context.Context, request domainGroup.JoinGroupWithLinkRequest) (response domainGroup.JoinGroupWithLinkResponse, err error) {
+	whatsapp.MustLogin(service.WaCli)
+
+	jid, err := service.WaCli.JoinGroupWithLink(request.Link)
+	if err != nil {
+		return
+	}
+	fmt.Printf("%+v\n", jid)
```
aldinokemal: i think we can remove this

aldinokemal (commented):
Please change this PR title

`feat: Add Join Group with Link`

aldinokemal:
Would you be able to assist with updating the readme.md file too?

src/internal/rest/group.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/gofiber/fiber/v2"
+)
+
+type Group struct {
+	Service domainGroup.IGroupService
+}
+
+func InitRestGroup(app *fiber.App, service domainGroup.IGroupService) Group {
+	rest := Group{Service: service}
+	app.Post("/group/join_with_link", rest.JoinGroupWithLink)
```
aldinokemal: I prefer `/group/join-with-link` instead of `/group/join_with_link`

src/services/group.go
tim-vandecasteele: Makes sense, I removed it.

src/internal/rest/group.go
tim-vandecasteele: Thanks, didn't know what the convention was as there weren't any links composed of multiple words yet, I changed it.

tim-vandecasteele:
Thanks for the feedback, I did the changes and added the URL in the readme, any more changes you want to see?

aldinokemal (approved):
Thank you for your contribution. I greatly appreciate it, and it looks good to me (LGTM).

Issue #81, opened 2023-07-04
Webhook "message" attribute is empty when sending from WhatsApp Web


Itay-Sharoni:
when using webhook (lets call it Bot)
if a user send a message to the Bot, then the webhook post:
{
  "audio": null,
  "contact": null,
  "document": null,
  "forwarded": null,
  **"from": "97254555555@s.whatsapp.net",**
  "image": null,
  "list": null,
  "live_location": null,
  "location": null,
  **"message": "This is a Test!!",**
  "message_id": "A91DB2EDE7D5B5DBEE63034A96B0A13D",
  "order": null,
  "sticker": null,
  "video": null,
  "view_once": null
}
this is good since i sent "This is a Test!!"

however, now same user send the exact same message but now he uses his WhatsApp Web on his PC
this time there are 2 differences, the "from" attribute look different and message attribute is empty this way

so end-users who interact with the bot must use their phone app instead the PC/ Web App
{
  "audio": null,
  "contact": null,
  "document": null,
  "forwarded": null,
  **"from": "97254555555.0:12@s.whatsapp.net in 97254555555@s.whatsapp.net",**
  "image": null,
  "list": null,
  "live_location": null,
  "location": null,
  **"message": ""**,
  "message_id": "C9C4663CA1BEB301A3",
  "order": null,
  "sticker": null,
  "video": null,
  "view_once": null
}

Itay-Sharoni:
another thing i noticed when i started with --debug

when a user send a message to the phone configured with go-whatsapp
if the end-user used his phone (android in this example) it looks like this (notice my text "Lala" it is in attribute "conversation") :

14:42:51.216 [Main INFO] Received message 0A3CB12A413152CFA0C9FEF694D1EDDB from 97254555555@s.whatsapp.net (pushname: Itay Sharoni, timestamp: 2023-07-04 14:42:51 +0000 UTC, type: text): conversation:"Lala"  messageContextInfo:{deviceListMetadata:{senderKeyHash:"\xa9\x82\xad\xa0?\xd9\tK\x88\x01"  senderTimestamp:1687936453  recipientKeyHash:"V\x98\xd78,kа҅"  recipientTimestamp:1688474298}  deviceListMetadataVersion:2}


but if the same end-user used WhatsApp Web or App for PC it looks like this (notice my text "aasdsd" it is in attribute "extendedTextMessage" and inside that the text is in "text") :

[...]

PR #82, opened 2023-07-05
add condition, if message received with ExtendedTextMessage, then use it, else use Conversation attribute

Itay-Sharoni:
in continue to issue: #81 
add condition, if message received with ExtendedTextMessage, then use it, else use Conversation attribute


Itay-Sharoni (commented):
I tested on my server and it works for both scenarios

aldinokemal:
so this is regarding your issue right?
https://github.com/aldinokemal/go-whatsapp-web-multidevice/issues/81

aldinokemal (approved):
Many thanks for the effort you've put forth.

PR #112, opened 2024-02-04
Updated quotedMessage to include original message ID for correlation

tcreswick:
I've updated this to include the message context when the incoming message is a reply. This means that the message ID of the message being replied-to is included, which is very helpful (otherwise there is no way to correlate a specific message).

In my application, I send a series of messages to the user, and save the corresponding message ID after sending. When the user replies to a specific message, I am able to reference the correlated ID and my application can then handle that in context.

### DIFF
--- a/whatsapp-api.service
+++ b/whatsapp-api.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=WhatsApp API Service
+After=network-online.target
+
+[Service]
+User=root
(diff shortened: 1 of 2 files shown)


whatsapp-api.service
aldinokemal, line 1: we don't need this

aldinokemal (commented):
can you provide in PR detail, before and after your changes?

PR #159, opened 2024-06-08
Enable users to add webhooks through a .env file

abdbbdii:
## Context
- Enable users to add webhooks through a .env file for easier deployments since websites like [Render](https://render.com/) don't let you write custom commands to compose Dockerized projects.


abdbbdii (commented):
All good

PR #181, opened 2024-08-27
Provide settings from environmental variables

abdbbdii:
## Context
- Enable users to add webhooks and other settings from environmental variables for easier deployments since websites like [Render](https://render.com/) don't let you write custom commands to compose Dockerized projects.


abdbbdii:
Hi, thank you for this awesome app.
Could you please consider adding this functionality natively? Currently, I have to manually copy and paste the functions to retrieve environment variables. Additionally, it doesn't update automatically because I fork the project first and then make changes. As a result, the deployed version stops working after a major update.
Regards.

Issue #184, opened 2024-08-31
⚡️ Add a Secret option to Webhook


LuizEdgar:
### What problem does your feature request solve?

There is no way to know who is calling my Webhook. 
I want to add a Secret to GoWhatsapp so it sends the secret when calling the Webhook and then I csan verify this was sent by the GoWhatsapp.

### Describe the solution you are proposing.

1. Add a new option like `--webhook-secret=abcdefghijklm12345` 
2. The secret `abcdefghijklm12345` is sent in a Header on the Webhook HTTPs request.
3. I check the `abcdefghijklm12345` on my side.

### Importance

I Need It

### Additional Context

- [ ] This missing feature presents as a bug.
- [X] This missing feature is forcing me to consider alternatives.

aldinokemal:
interesting, will prioritize this

sohepalslamat:
Is there any news about this case? I need to receive messages.

aldinokemal:
https://github.com/aldinokemal/go-whatsapp-web-multidevice/releases/tag/v4.18.0

LuizEdgar:
Amazing! Thanks! :)

PR #187, opened 2024-09-13
add webhook secret option

mr687:
## Context
https://github.com/aldinokemal/go-whatsapp-web-multidevice/issues/184

1. Add new option `--webhook-secret="super-secret-key"`
2. Every webhook will send extra header `X-Hub-Signature-256: sha256={signature}` to the target client
3. Client will verify the signature

#  Example Usage
1. Start whatsapp service for example using docker.
```bash
docker run --detach \
--publish=3000:3000 \
--name=whatsapp \
--restart=always \
--volume=$(docker volume create --name=whatsapp):/app/storages \
aldinokemal2104/go-whatsapp-web-multidevice \
--webhook="http://yourwebhook.site/webhook/whatsapp" \
--webhook-secret="super-secret-key"
```

2. Example webhook content:
```json
// Headers
{
  "X-Hub-Signature-256": "sha256=1790a93ec78e7d0294884cd983c063756f8a068c53b1123600d0fe8678f7c830",
  "Content-Type": "application/json",
  "Lontent-Length": "406",
  "User-Agent": "Go-http-client/1.1",
  ...
}
```

```json
// Body
{
  "audio": null,
  "contact": null,
  "document": null,
  "forwarded": false,
  "from": "xxx@s.whatsapp.net in xxxx@s.whatsapp.net",
  "image": null,
  ....
}
```

3. Verify webhook signature.
```javascript
const crypto = require('node:crypto')
const express = require('express')

const app = express()
const secretKey = "super-secret-key"

app.post("/webhook/whatsapp", async (req, res) => {
  const signature = req.headers['X-Hub-Signature-256'] || ""
  const body = req.body

  const expectedHash = crypto.createHmac("sha256", secretKey)
    .update(JSON.stringify(body))
    .digest("hex")
  const actualHash = signature.slice("sha256=".length)
  const isValidSignature = expectedHash === actualHash

  if (!isValidSignature) {
    throw new BadRequestException('Signature invalid!');
  }

  // DOING SOMETHING GREAT...

  return res.send("ok")
})
```


mr687:
Hi bang @aldinokemal, can you please approve this PR as I need this webhook secret for my project. thanks.

src/config/settings.go
```diff
@@ -21,6 +21,7 @@ var (
 
 	WhatsappAutoReplyMessage    string
 	WhatsappWebhook             string
+	WhatsappWebhookSecret       string
```
aldinokemal: I think we can set default value here "secret"

src/pkg/whatsapp/whatsapp.go
```diff
@@ -400,7 +413,22 @@ func forwardToWebhook(evt *events.Message) error {
 	if err != nil {
 		return pkgError.WebhookError(fmt.Sprintf("error when create http object %v", err))
 	}
+
+	var secretKey []byte
+	if config.WhatsappWebhookSecret != "" {
+		secretKey = []byte(config.WhatsappWebhookSecret)
+	} else {
+		secretKey = []byte("anything")
```
aldinokemal: no need if else, use config.WhatsappWebhookSecret with default value if not provided

aldinokemal (commented):
Thank for your efforts for providing HMAC into webhook header, just left some comment
Could you provided webhook capture?

src/pkg/whatsapp/whatsapp.go
```diff
@@ -400,7 +413,22 @@ func forwardToWebhook(evt *events.Message) error {
...
+	}
+
+	signature, err := getMessageDigestOrSignature(postBody, secretKey)
+	if err != nil {
+		return pkgError.WebhookError(fmt.Sprintf("error when create signature %v", err))
+	}
+
 	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("X-Hub-Signature-256", fmt.Sprintf("sha256=%s", signature))
```
aldinokemal, line 424: Change the name to `X-Signature-256` as this isn't a github webhook. Alternatively, you can provide me with a reference website that uses `X-Hub-Signature-256` instead of github.

mr687: `X-Hub-Signature-256` is based on the standard used by Facebook for Whatsapp webhooks. I wanted to keep consistency with their implementation for signature verification.

Let me know if you'd still prefer the header to be changed to `X-Signature-256`. I'm happy to adjust it.

Ref: https://developers.facebook.com/docs/messenger-platform/webhooks/#validate-payloads

mr687:
> Thank for your efforts for providing HMAC into webhook header, just left some comment Could you provided webhook capture?

Oh I missed that, okay I’ve updated the description.

aldinokemal (approved):
LGTM

PR #203, opened 2024-10-10
HTTPS & HTTP Problem Axios Request

decoderid:
## Context
Instead of using .AppHost, there will be issues when using a domain, for example, Cloudflare HTTPS, while on the host side using HTTP requests.

##  Test Results

### DIFF
--- a/src/views/index.html
+++ b/src/views/index.html
@@ -120,7 +120,7 @@
     }
 
     window.http = axios.create({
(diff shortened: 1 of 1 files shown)


aldinokemal:
That makes sense. Thank you for your PR.
LGTM

aldinokemal (approved):

PR #225, opened 2025-01-07
Refactored WebSocket URL construction into a separate constructWebSoc…

codejq:
…ketURL method and updated login/logout API paths to use relative URLs instead of absolute paths.

## Context

##  Test Results


aldinokemal (approved):
LGTM
Thanks for your effort

PR #301, opened 2025-06-19
IPv6 support.

zlenner:
## Context
I'm hosting my app on railway, which supports [private networking only on IPv6](https://docs.railway.com/guides/private-networking).

The WhatsApp API was only listening for IPv4, so I had to update fiber's "Network" variable.

Default is tcp4. tcp6 listens on only IPv6 connections, and [tcp listens on both](https://github.com/gofiber/fiber/blob/3b2af61152d0dc9bb75bd9451e506274266b638e/listen.go#L69).

Will not break anything AFAIK.


zlenner:
I changed 14 characters- I'm not reading whatever problem sEmAntIc pR has with it.

PR #305, opened 2025-06-23
Boolean env var option for enabling calls to webhook on self message

TheProcedural:
## Context
Added the option to enable calls to webhook on self messages. This is deactivated by default like it currently works prior to this feature, but now with the possibility of enabling with the boolean env var CALL_WEBHOOK_ON_MESSAGE_TO_SELF

##  Test Results
I've run my own version for several days, still running it right now to create a more complete client like experience.


aldinokemal:
I have removed code isFromMySelf

TheProcedural:
> I have removed code isFromMySelf

what?

PR #414, opened 2025-09-12
Add Typescript API lib and generator

zolotarev:
# Add Typescript API lib and generator

## Context

- Just added API-file for TS and generator bash-script


aldinokemal:
for api client can generate using openapi generator

PR #427, opened 2025-09-28
Support X-Forwarded Headers for Reverse Proxy Compatibility in Login Endpoint

wurin7i:
Support X-Forwarded Headers for Reverse Proxy Compatibility in Login Endpoint

## Context

This PR updates the /app/login endpoint to properly handle X-Forwarded-Host, X-Forwarded-Proto, and X-Forwarded-Port headers. This change ensures correct URL generation when the application is deployed behind a reverse proxy (such as Kong, Envoy, Traefik, or HAProxy). If these headers are present, the app uses their values for host, protocol, and port; otherwise, it falls back to the default request values. This improves compatibility and reliability in proxied environments.

## Test Results
image

### DIFF
--- a/src/ui/rest/app.go
+++ b/src/ui/rest/app.go
@@ -30,12 +30,22 @@ func (handler *App) Login(c *fiber.Ctx) error {
 	response, err := handler.Service.Login(c.UserContext())
 	utils.PanicIfNeeded(err)
 
+	host := c.Get("X-Forwarded-Host")
+	if host == "" {
+		host = c.Hostname()
+	}
+
+	protocol := c.Get("X-Forwarded-Proto")
+	if protocol == "" {
+		protocol = c.Protocol()
+	}
+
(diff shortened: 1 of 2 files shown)


src/ui/rest/app.go
aldinokemal, line 1: Thank you for your effort, mas @wurin7i, in creating this PR. Why is the reverse proxy only for app.go?

wurin7i: The reverse proxy header handling was added in `app.go` because the `/app/login` endpoint generates a QR link that needs to reflect the correct external URL when accessed behind a proxy. Other endpoints do not generate URLs for clients, so they do not require this logic.

aldinokemal:
After doing some research on docs, I think we can just set `EnableTrustedProxyCheck: true`.

image

I'll add it in the next version, thanks for your effort. I'll close this for now.

PR #453, opened 2025-11-05
feat(schedule): add scheduled messaging CRUD (domain/usecase/repo/REST) and ScheduleManager UI with filters + manual refresh

mdelrosario92:
feat(schedule): scheduled messaging CRUD + UI manager

Overview
- Adds scheduling across domain/usecase/repository with lifecycle: pending → sending → sent/failed
- REST endpoints for CRUD + run-now
- Vue Schedule Manager to list, filter, create, edit, run-now, delete
- Start-up recovery for stuck items (sending → pending at boot if stale)

Why
- Reliable deferred delivery for campaigns, reminders and off-hours sending
- Operational visibility: see/retry failures, edit pending messages, and dispatch early when needed

Endpoints
- GET /schedule/messages?statuses=pending,failed&limit=100&offset=0
- POST /schedule/messages
- PUT /schedule/messages/{id}
- DELETE /schedule/messages/{id}
- POST /schedule/messages/{id}/run

Docs
- OpenAPI updated for all new paths and schemas (ScheduledMessage, ScheduleMessagePayload)

UI
- New ScheduleManager card in the dashboard
- Filters with explicit Refresh
- Create/Edit modal validates schedule lead time (≥5s) and ≤30 days

Screenshots
- Schedule Card
  - schedule-card
- Schedule List Modal
  - schedule-modal
- Create Form
  - schedule-create-form

Behavioral details
- Only pending messages are eligible for run-now
- Update/Delete allowed for pending/failed only (guards in repo)
- Dispatcher continues to run from send usecase; RunNow uses same path
- Boot-time ResetStuck moves long-running sending → pending


mdelrosario92:
Thanks for the thorough walkthrough and checks — a few clarifications and updates made in response:

- OpenAPI: Added full documentation for the new schedule endpoints and schemas, and extended /send/message to include `schedule_at` in the request and `scheduled_id`/`scheduled_for` in the response.
- Screenshots: Attached UI screenshots (card, list modal, create form) to the PR description for easier visual review.
- HTTP status on create: kept 200 for now to remain consistent with the existing API pattern across the repo (most create endpoints return 200 with a `ResponseData` envelope). If maintainers prefer 201 for POST /schedule/messages, happy to switch in a follow‑up commit.
- Concurrency and transitions: `MarkProcessing` uses a single conditional UPDATE (WHERE status=pending) to atomically acquire work; `ResetStuck` on boot returns long‑running `sending` back to `pending`. `RunNow` performs MarkProcessing → Send → MarkSent/MarkFailed, with error truncation to keep logs tidy.
- Indexes/timestamps: scheduled_messages table already has status and schedule_at indexes in the existing migration; all stored times (including schedule_at) are normalized to UTC. UI renders in local time.
- UI behavior: filters are explicit (checkboxes + Refresh) to make the interaction clearer; form enforces lead time ≥5s and ≤30 days, validates message length, and composes JID correctly for user/group/newsletter/status.

If there’s a strong preference for (a) 201 Created on schedule create, (b) a GET /schedule/messages/{id} convenience endpoint, or (c) a short repo test around List/Update/Delete guards, I can add those in this PR.

mdelrosario92:
Added a small enhancement based on the review thread:

- New endpoint: GET /schedule/messages/{id}
  - REST handler + OpenAPI spec documented
  - Mirrors List payload, returns a single ScheduledMessage (200)

This is optional but handy for clients retrieving a newly-created or updated item by ID without re-listing.

mdelrosario92:
Addressed the latest nits:

- infra: MarkSent/MarkFailed now check RowsAffected and return sql.ErrNoRows when 0 to avoid silent no-ops.
- usecase: truncateError now truncates at rune boundaries to preserve valid UTF‑8.
- UI: formatStatus no longer echoes raw status in v-html; unrecognized statuses map to a fixed "Unknown" label to avoid XSS risk.

Thanks for the sharp catches.

mdelrosario92:
Hey @aldinokemal, what do you think? Would love your input. Thanks!

aldinokemal:
Hey, sorry, just finished reviewing. This is a great PR! But we should also think about adding the option to schedule sending media messages too.

src/cmd/root.go
```diff
@@ -261,7 +274,8 @@ func initApp() {
 	// Usecase
 	appUsecase = usecase.NewAppService(chatStorageRepo)
 	chatUsecase = usecase.NewChatService(chatStorageRepo)
-	sendUsecase = usecase.NewSendService(appUsecase, chatStorageRepo)
+	sendUsecase = usecase.NewSendService(appUsecase, chatStorageRepo, scheduleRepo)
```
aldinokemal, line 277: 🔴 **Blocker — this head does not compile.** `NewSendService` at this PR's head still accepts only `(appService, chatStorageRepo)`, but this call passes a third `scheduleRepo` argument. The same incomplete integration appears in `serviceSchedule`: it references `MessageRequest.ScheduleAt` and `GenericResponse.ScheduledID`, neither of which exists at this head. Please include the send-domain and `SendText` scheduling changes in the PR and run `go test ./...` on the exact head.

src/cmd/root.go
```diff
@@ -250,6 +254,15 @@ func initApp() {
 	chatStorageRepo = chatstorage.NewStorageRepository(chatStorageDB)
```
aldinokemal, line 254: 🔴 **Blocker — `scheduled_messages` is never created.** This PR constructs the repository and immediately calls `ResetStuck`, but none of the changed files adds a schema migration/table creation for `scheduled_messages`. Startup will log `no such table`, and every schedule CRUD/dispatch operation will fail afterward. Add the migration to the existing chat-storage migration sequence and fail initialization if it cannot be applied.

aldinokemal (commented):
Reviewed the scheduling lifecycle, repository state transitions, REST wiring, send-path integration, startup recovery, prior review threads, and available CI status.

This PR is currently blocked before runtime testing because the head does not compile and the persistence schema is incomplete. In addition to the inline findings, `src/usecase/schedule.go` initializes `domainSend.MessageRequest.ScheduleAt` and reads `domainSend.GenericResponse.ScheduledID`, but the head versions of those types do not define either field. The head `SendText` implementation also always sends immediately; it contains no branch that persists a future request into `scheduled_messages`.

Please add the missing send-domain fields and scheduling branch as one coherent change, create the `scheduled_messages` migration before constructing the repository, and add tests for: future request queues without sending, due dispatch transitions `pending → sending → sent`, send failure becomes `failed`, concurrent/manual dispatch acquires only once, and restart recovery. No pull-request workflow runs were available for this head SHA, and GitHub currently reports the PR as non-mergeable.

aldinokemal:
not relevant anymore, we migrate UI into https://github.com/aldinokemal/gowa-ui

Issue #459, opened 2025-11-10
🐞 Webhook message_edited event returns new message ID instead of original message ID


muneebanwar33:
### What Happened:
When a message is edited in WhatsApp, the webhook `message_edited` event returns a **new message ID** instead of the **original message ID** that was edited. This makes it impossible to track which message was actually edited.

**Current webhook payload when message is edited:**
```json
{
  "action": "message_edited",
  "chat_id": "120363422905862684",
  "edited_text": "12a",
  "from": "xxxxxxxxxxxxx:4@s.whatsapp.net in 120363422905862684@g.us",
  "from_lid": "xxxxxxxxxxxxx:4@lid",
  "message": {
    "text": "",
    "id": "3BD618DD7A77628E97B0",
    "replied_id": "",
    "quoted_message": ""
  },
  "pushname": "Test User",
  "sender_id": "xxxxxxxxxxxxx",
  "timestamp": "2025-11-10T01:33:34Z"
}
```

The `message.id` field shows `3BD618DD7A77628E97B0`, but the original message had ID `3BEE9038C685758CFEC4`. There's no way to know which message was edited.

### What I Expected:
The webhook should include the **original message ID** so I can update the correct message in my database. Either:

**Option 1 - Include original ID in existing field:**
```json
{
  "action": "message_edited",
  "message": {
    "id": "3BEE9038C685758CFEC4",
    "text": "12a"
  }
}
```

**Option 2 - Add a new field for original message ID:**
```json
{
  "action": "message_edited",
  "original_message_id": "3BEE9038C685758CFEC4",
  "edited_text": "12a",
  "message": {
    "id": "3BD618DD7A77628E97B0"
  }
}
```

### Reproduction Steps

```bash
Step 1: Send a message
Send a message to any WhatsApp chat/group via the API or manually

**Original message sent - Webhook received:**

{
  "chat_id": "120363422905862684",
  "from": "xxxxxxxxxxxxx:4@s.whatsapp.net in 120363422905862684@g.us",
  "from_lid": "xxxxxxxxxxxxx:4@lid",
  "message": {
    "text": "12",
    "id": "3BEE9038C685758CFEC4",
    "replied_id": "",
    "quoted_message": ""
  },
  "pushname": "Test User",
  "sender_id": "xxxxxxxxxxxxx",
  "timestamp": "2025-11-10T01:33:17Z"
}


### Step 2: Note the message ID
[...]

PR #460, opened 2025-11-10
feat: add original message ID and timestamp to message edit events

muneebanwar33:
Previously, when a message was edited, the webhook payload only included the new edited text but not the original message ID. This made it impossible to track which message was modified, as the payload contained a new event ID instead of referencing the original message.

This change extracts the original message metadata from the protocol message using GetKey(), similar to how message revokes are handled.

Changes:
- Add original_message_id field to track which message was edited
- Add original_from_me to indicate message ownership
- Add original_chat to include the chat JID
- Add edit_timestamp to capture when the edit occurred
- Update webhook payload documentation with new fields

The webhook payload now properly links edit events to their original messages, enabling applications to maintain accurate message history and display edits correctly.

# PR Title

## Context

## Test Results

### DIFF
--- a/docs/webhook-payload.md
+++ b/docs/webhook-payload.md
@@ -511,6 +511,10 @@ Triggered when users are demoted from admin.
   "action": "message_edited",
   "chat_id": "6289XXXXXXXXX",
   "edited_text": "hhhiawww",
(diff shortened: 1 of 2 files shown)


muneebanwar33:
Resolves #459

aldinokemal:
thanks for your effor, will add on next release with simpler changes

PR #486, opened 2025-12-04
feat: Added webhook forward filters

jayeeliu:
# Added webhook forward filters
 
## Context
Many times, the receiver doesn't need all messages, so added webhook forward filters.

1. message type: Only set type messages are forwarded. support all, text, image, video, document, sticker, contact, location, live_location, text_edited. default all.
2. receipt: When set to false, the event of message.ack is not forwarded. default true.
3. delete: When set to false, the event of delete message from another device is not forwarded. default true.


aldinokemal:
this should be done on client side not from this app

PR #487, opened 2025-12-06
fix: use config.AppOs directly to resolve connection/pairing issues

sonarta:
## Description
This PR addresses connection and pairing issues (e.g., "Unable to link device" or infinite loading during pairing) caused by the non-standard OS name format sent during the handshake.

## Changes
- Modified src/infrastructure/whatsapp/init.go: Removed the auto-appending of AppVersion to DeviceProps.Os. Now it strictly uses config.AppOs.
- Modified src/usecase/app.go: Updated LoginWithCode to use config.AppOs instead of the hardcoded "Chrome (Linux)" string, ensuring consistency between session properties and pairing requests.

## Why
Some WhatsApp anti-bot measures seem to flag the custom OS string format (e.g., "Chrome v7.10.1"). By allowing the user to specify a clean OS name (e.g., "Chrome" or "Mac OS") via the config, these connection issues are resolved.

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -1,7 +1,7 @@
 openapi: "3.0.0"
 info:
   title: WhatsApp API MultiDevice
-  version: 6.12.0
+  version: 6.13.0
   description: This API is used for sending whatsapp via API
 servers:
   - url: http://localhost:3000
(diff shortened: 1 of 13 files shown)


aldinokemal:
Thanks for the PR and for digging into the pairing issue.\n\nAfter reviewing this more closely, I’m going to close this one for now. The main issue is that  only sets the client type/platform ID, while the  passed to  still needs to be a valid browser-style string in the  format.\n\nIn this repo,  is effectively just an internal label and may contain values that are not valid for WhatsApp pairing display names. Because of that, changing  is not something I can safely merge as a fix.\n\nIf you want to revisit this, the safer direction would be to keep a valid pairing display name for  (for example ) and, if needed, introduce a separate config specifically for that field rather than reusing .\n\nAppreciate the contribution.

aldinokemal:
Small clarification because my previous comment lost some formatting:

Thanks again for the PR and for investigating the pairing issue.

I’m closing this for now because `PairClientChrome` only sets the client type / platform ID. The `clientDisplayName` passed to `PairPhone()` still needs to be a valid browser-style string in the `Browser (OS)` format.

In this repo, `AppOs` is effectively just an internal label and may contain values that are not valid for WhatsApp pairing display names. Because of that, changing `PairPhone(..., config.AppOs)` is not something I can safely merge as a fix.

If you want to revisit this later, the safer approach would be:
- keep a valid pairing display name for `PairPhone()` such as `Chrome (Linux)`
- if configurability is needed, introduce a separate config specifically for that pairing display field
- avoid reusing `AppOs` for the pairing display name

Appreciate the contribution.

PR #506, opened 2025-12-29
Fix webhook event filtering and refactor webhook broadcasting

ramon-victor:
This PR fixes a critical bug where the `WHATSAPP_WEBHOOK_EVENTS` configuration was not working correctly. The application was failing to filter events properly because the internal event names passed to the whitelist checker did not match the expected configuration keys (e.g., passing "message event" instead of "message"). Consequently, valid events were being filtered out when a whitelist was active.

**Key Changes:**
- **Fix Event Naming**: Updated all event handlers (`message`, `receipt`, `delete`) to use standard event keys (e.g., `message`, `message.ack`, `message.delete`) that match the configuration expectations.
- **Fix Group Event Filtering**: Updated the group participant event handler to respect the `WHATSAPP_WEBHOOK_EVENTS` whitelist (previously it bypassed the check entirely).
- **Refactor Webhook Broadcasting**:
  - Centralized webhook logic into a clean `BroadcastWebhookEvent` helper in `src/infrastructure/whatsapp/webhook_forward.go`.
  - Decomposed complex logic into smaller, testable functions (`shouldForwardEvent`, `dispatchToWebhooks`, `handleBroadcastResults`).

### DIFF
--- a/src/infrastructure/whatsapp/event_delete.go
+++ b/src/infrastructure/whatsapp/event_delete.go
@@ -16,7 +16,7 @@ func forwardDeleteToWebhook(ctx context.Context, evt *events.DeleteForMe, messag
 		return err
 	}
 
-	return forwardPayloadToConfiguredWebhooks(ctx, payload, "delete event")
+	return BroadcastWebhookEvent(ctx, "message.delete", payload)
 }
 
 // createDeletePayload creates a webhook payload for delete events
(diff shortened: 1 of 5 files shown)


aldinokemal:
already in v8.1

ramon-victor:
> already in v8.1

Yes, but this PR is precisely to fix WHATSAPP_WEBHOOK_EVENTS.
Even if you define all events (for example, exactly as in .env.example), no events are sent to the webhook at all, as I already explained in the Pull Request description.

PR #522, opened 2026-01-07
feat: add healthcheck endpoint

icl00ud:
This PR adds a simple /health endpoint that returns 'OK' to verify the service status.


aldinokemal:
Hey! Nice addition 👍

A couple of thoughts:

**Health endpoint path**
The `/health` endpoint is registered at root while other routes use `config.AppBasePath`. This means if someone deploys with `--base-path="/gowa"`, health will still be at `/health` instead of `/gowa/health`. Totally fine if intentional (k8s probes often expect fixed paths), might just be worth a quick comment noting that it's deliberately at root.

**What "healthy" means**
The current check verifies the store is initialized, which works great for basic liveness. Just noting that it won't reflect actual WhatsApp connection status—might be worth mentioning in the comment that this is an initialization check rather than connectivity check, so future readers know what to expect.

Neither of these are blockers, just things that caught my eye. 🙂

icl00ud:
> Hey! Nice addition 👍
> 
> A couple of thoughts:
> 
> **Health endpoint path** The `/health` endpoint is registered at root while other routes use `config.AppBasePath`. This means if someone deploys with `--base-path="/gowa"`, health will still be at `/health` instead of `/gowa/health`. Totally fine if intentional (k8s probes often expect fixed paths), might just be worth a quick comment noting that it's deliberately at root.
> 
> **What "healthy" means** The current check verifies the store is initialized, which works great for basic liveness. Just noting that it won't reflect actual WhatsApp connection status—might be worth mentioning in the comment that this is an initialization check rather than connectivity check, so future readers know what to expect.
> 
> Neither of these are blockers, just things that caught my eye. 🙂

Hi, aldino!

The changes has been made! Thank you for the feedback :)

PR #532, opened 2026-01-09
feat - add /docs route for openapi swagger view

ilanbenb:
# PR Title

Add /docs url for ease of using the swagger file

## Context
image

## Test Results

I started it locally and called a few endpoints

I decided to implement with fiber-swagger ( new dependency ) since it seems like more of a standard.
Could have gone with a zero-dependency route as well. Let me know what you prefer.

  Option 1: Zero-dependency (CDN-based)
  - Add a simple HTML file that loads Swagger UI from CDN
  - Serve your existing openapi.yaml at /docs/openapi.yaml
  - No new Go dependencies needed
  - Lightweight, ~20 lines of code
  
  Option 2: Use fiber-swagger package ( choosed )
  - Add github.com/arsmn/fiber-swagger/v2 dependency
  - More "standard" Go approach
  - Slightly more setup, but provides middleware


ilanbenb:
@aldinokemal WDYT?

aldinokemal (commented):
Reviewed the Swagger route, base-path/auth placement, module changes, runtime packaging, prior discussion, and available CI status.

🔴 **Blocker — `src/cmd/rest.go` (Swagger spec serving):** the existing unresolved review thread is still valid. The production Docker image uses `WORKDIR /app` and copies only `/app/whatsapp`; it does not contain `docs/openapi.yaml`. Therefore `../docs/openapi.yaml` resolves to `/docs/openapi.yaml`, and `/openapi.yaml` will fail in the deployed container even if the Swagger UI HTML itself loads.

Please embed the spec into the binary (preferred), or explicitly copy it into the runtime image and serve it from a deterministic path. Add a smoke test against the built container that verifies both `/docs/` and `/openapi.yaml` under an empty and non-empty `AppBasePath`.

The branch is also 191 commits behind `main`, GitHub currently reports it as non-mergeable, and no pull-request workflow runs were available for the head SHA. Please rebase and rerun build/tests after resolving the blocker. I did not duplicate the inline comment because the exact defect already has an unresolved thread.

src/go.mod
```diff
@@ -41,11 +44,17 @@ require (
 	github.com/elliotchance/orderedmap/v3 v3.1.0 // indirect
 	github.com/fasthttp/websocket v1.5.12 // indirect
 	github.com/fsnotify/fsnotify v1.9.0 // indirect
+	github.com/go-openapi/jsonpointer v0.19.5 // indirect
+	github.com/go-openapi/jsonreference v0.19.6 // indirect
+	github.com/go-openapi/spec v0.20.4 // indirect
+	github.com/go-openapi/swag v0.19.15 // indirect
 	github.com/go-viper/mapstructure/v2 v2.4.0 // indirect
```
aldinokemal, line 51: **Rebase blocker:** current `main` now uses `github.com/gofiber/fiber/v3 v3.4.0`, but `github.com/gofiber/swagger v1.1.1` declares a dependency on `github.com/gofiber/fiber/v2 v2.52.6`. Its handler type therefore cannot be registered directly on the current Fiber v3 router after rebasing. Please replace this with a Fiber-v3-compatible docs implementation (or a small self-contained handler) instead of reintroducing the v2 stack.

aldinokemal (changes requested):
Re-reviewed this against current `main` and the production image. This is still not merge-safe.

Two blockers remain:

1. The existing unresolved Swagger-spec serving defect is still valid: the PR serves `../docs/openapi.yaml` from the runtime working directory, but the production image copies only `/app/whatsapp` plus the entrypoint into `WORKDIR /app`. The spec is therefore absent in production, so `/openapi.yaml` (and consequently Swagger UI) cannot load the document. Embed the spec into the binary or otherwise package it at a deterministic runtime path, and register the routes through the current `apiGroup`/base-path structure.

2. Current `main` has migrated to `github.com/gofiber/fiber/v3 v3.4.0`, while this PR adds `github.com/gofiber/swagger v1.1.1`; that module itself requires `github.com/gofiber/fiber/v2`. After the required rebase, its v2 handler is not compatible with the current Fiber v3 router. Please use a Fiber-v3-compatible Swagger approach (or a small self-contained docs handler) rather than carrying the v2 adapter back into the module graph.

GitHub currently reports the PR as non-mergeable, and there are no pull-request workflow runs for head `8bd7a7715b15e461062aa076b8ee7cf34fb866fd` (only the CodeRabbit commit status). After rebasing and fixing the two issues, please add a smoke test covering `/docs` + `/openapi.yaml` both with empty and non-empty `AppBasePath`, and verify the production container path.

src/cmd/rest.go
```diff
@@ -97,6 +98,15 @@ func restServer(_ *cobra.Command, _ []string) {
 		apiGroup = app.Group(config.AppBasePath)
 	}
 
+	// Swagger UI documentation
+	app.Static(config.AppBasePath+"/openapi.yaml", "../docs/openapi.yaml")
```
aldinokemal, line 102: **P1 — this spec path does not exist in the production image.** Current production `docker/golang.Dockerfile` builds from `src/`, then the runtime stage copies only `/app/whatsapp` and `/entrypoint.sh` into `WORKDIR /app`. This line therefore resolves `../docs/openapi.yaml` to a file that is never shipped, so `/openapi.yaml` fails in the deployed container and Swagger UI cannot load its contract. Embed `docs/openapi.yaml` into the binary (preferred) or explicitly package it at a deterministic runtime path, then cover the built image with `/docs` + `/openapi.yaml` smoke tests for empty and non-empty `APP_BASE_PATH`.

src/go.mod
```diff
@@ -41,11 +44,17 @@ require (
...
+	github.com/gofiber/swagger v1.1.1 // indirect
```
aldinokemal, line 52: **P1 — this Swagger adapter is incompatible with current-main Fiber.** Current `main` is on `github.com/gofiber/fiber/v3 v3.4.0`, while `github.com/gofiber/swagger v1.1.1` itself requires `github.com/gofiber/fiber/v2 v2.52.6`. After the required rebase, its handler/router types cannot be wired into the Fiber v3 `fiber.Router` used by `rest.go`. Please use a Fiber-v3-compatible approach or a small self-contained docs handler instead of carrying the v2 adapter into the module graph.

aldinokemal (changes requested):
Re-review against current `main` (`b2a57a2062d5dcebcb70c6dfd05c8aec5b613868`) confirms this PR is still not merge-safe.

Two blockers remain:

1. The OpenAPI document is served from a runtime-relative filesystem path, but the production image contains only `/app/whatsapp` plus the entrypoint. The Swagger UI can render while `/openapi.yaml` fails because `docs/openapi.yaml` is not present. Please embed the spec into the binary (preferred) or package it explicitly at a deterministic path, and add a production-image smoke test for `/docs` + `/openapi.yaml` with both empty and non-empty `APP_BASE_PATH`.

2. Current `main` uses `github.com/gofiber/fiber/v3 v3.4.0`. This branch adds `github.com/gofiber/swagger v1.1.1`, whose own module requires Fiber v2. A required rebase therefore cannot carry this handler forward as-is. Please use a Fiber-v3-compatible/self-contained docs handler rather than reintroducing the v2 router type/module graph.

GitHub currently reports the PR non-mergeable and there are no PR-triggered workflow runs for head `8bd7a7715b15e461062aa076b8ee7cf34fb866fd`. Rebase first, preserve current-main auth/base-path routing, then run the normal build/test checks and the container smoke test.

PR #561, opened 2026-02-03
fix: HTTP request context canceling WhatsApp connection prematurely

iamsantto:
# Fix HTTP request context canceling WhatsApp connection prematurely

## Context
- Fixed a bug where the HTTP request context was canceling the WhatsApp WebSocket connection before users could scan the QR code
- The original code passed c.UserContext() to the WhatsApp client, which gets canceled when the HTTP response is sent (~26ms), making QR code scanning impossible
- Now uses context.Background() with appropriate timeouts so the WhatsApp connection stays alive independently of the HTTP request lifecycle

## Changes
src/ui/rest/app.go: Use a background context with 2-minute timeout instead of request context in the Login handler
src/usecase/app.go: Use context.Background() for the QR channel so it remains open for pairing

## Test Results
Built successfully with go build
QR code login flow now allows sufficient time for users to scan the code and complete pairing

### DIFF
--- a/src/usecase/app.go
+++ b/src/usecase/app.go
@@ -48,9 +48,11 @@ func (service *serviceApp) Login(ctx context.Context, deviceID string) (response
 	// Disconnect first to ensure QR flow starts cleanly.
 	client.Disconnect()
 
-	// Use a background context for the QR channel so it stays alive after the HTTP response is sent.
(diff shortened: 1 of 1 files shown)


aldinokemal:
yes correct, already fixed

PR #568, opened 2026-02-08
feat: add endpoint to retrieve message by ID with optional phone veri…

siddiqar:
…fication

# PR Title

## Context

## Test Results


aldinokemal:
I’m not convinced we should merge this PR in its current form.

The main concern is that this endpoint reads from our local chat storage / SQLite, while SQLite is **not the source of truth** for WhatsApp messages. Because of that, `GET /message/{message_id}` feels misleading: it sounds like an authoritative “get message by id” API, but in practice it only returns a locally persisted record if that record happens to exist.

That creates a few problems:

- the API semantics are stronger than the actual behavior
- the response depends on local storage completeness/retention
- users may assume this reflects WhatsApp state, when it does not

The optional `phone` verification helps validate chat ownership, but it does not solve the core issue that the data still comes from local cache/storage rather than WhatsApp itself.

Because of that, I don’t think this should be added as a general public endpoint in its current shape.

If there is a real use case for this, I think it needs to be reframed more explicitly as a **local cached message lookup**, for example:
- clearly documented as storage-backed only
- named/scoped in a way that does not imply source-of-truth semantics
- possibly exposed only for debug/admin/internal use cases

So for now, I’d prefer we do **not merge** this PR as-is.

aldinokemal:
Closing this for now — I don't think we should merge it in the current form because the endpoint relies on local SQLite/cache rather than a real source-of-truth message lookup.

PR #584, opened 2026-02-22
fix: set sequence write database

doko89:
# PR Title

## Context

## Test Results

### DIFF
--- a/src/config/settings.go
+++ b/src/config/settings.go
@@ -23,7 +23,7 @@ var (
 	PathMedia     = "statics/media"
 	PathStorages  = "storages"
 
-	DBURI     = "file:storages/whatsapp.db?_foreign_keys=on"
+	DBURI     = "file:storages/whatsapp.db?_journal_mode=WAL&_foreign_keys=on"
 	DBKeysURI = ""
 
 	WhatsappAutoReplyMessage          string


aldinokemal:
## Review: Impact Analysis of `_journal_mode=WAL`

Good intent — WAL mode improves SQLite concurrency by allowing readers and writers to operate simultaneously. The chat storage DB (`chatstorage.db`) already uses WAL mode, so this aligns the main DB with that pattern. ✅

However, there are a few things to address before merging:

### 1. ⚠️ Cleanup Code — WAL Sidecar Files Not Deleted

With WAL mode enabled, SQLite creates two additional sidecar files:
- `whatsapp.db-wal`
- `whatsapp.db-shm`

The `CleanupDatabase()` function in `src/infrastructure/whatsapp/cleanup.go` only removes `whatsapp.db` but **not** the WAL/SHM sidecar files. This could leave stale files on disk after logout/cleanup.

**Suggested fix** — add after the main `os.Remove(dbPath)` call:

```go
// Remove WAL sidecar files
os.Remove(dbPath + "-wal")
os.Remove(dbPath + "-shm")
```

Same fix should be applied for the `keysDB` file cleanup path in the same function.

### 2. ⚠️ `.env.example` Not Updated

Users who override `DB_URI` via `.env` won't benefit from this change since env vars completely replace the default. The `.env.example` should be updated to include WAL:

```
DB_URI="file:storages/whatsapp.db?_journal_mode=WAL&_foreign_keys=on"
```

### 3. 💡 (Nice-to-have) Consistent Approach

The chat storage DB handles WAL programmatically via a dedicated config flag (`ChatStorageEnableWAL`) and builds the connection string in `initChatStorage()`. Consider adopting the same approach for the main DB — use a config flag like `DBEnableWAL` and inject `_journal_mode=WAL` in code. This ensures WAL is applied regardless of how the user configures `DB_URI`.

PR #641, opened 2026-04-18
Expose referral metadata in chat message API

juliomuhlbauer:
## Summary
- Add `referral_metadata` to the chat message API payload
- Map the stored referral JSON into `GetChatMessages` responses
- Document the new field in the OpenAPI schema

## Verification
- `go test ./usecase ./domains/chat` *(not run here because `go` is unavailable in this environment)*

### DIFF
--- a/src/usecase/chat.go
+++ b/src/usecase/chat.go
@@ -165,19 +165,20 @@ func (service serviceChat) GetChatMessages(ctx context.Context, request domainCh
 	messageInfos := make([]domainChat.MessageInfo, 0, len(messages))
 	for _, message := range messages {
 		messageInfo := domainChat.MessageInfo{
-			ID:           message.ID,
-			ChatJID:      message.ChatJID,
-			SenderJID:    message.Sender,
-			Content:      message.Content,
-			Timestamp:    message.Timestamp.Format(time.RFC3339),
-			IsFromMe:     message.IsFromMe,
-			MediaType:    message.MediaType,
-			CallMetadata: message.CallMetadata,
-			Filename:     message.Filename,
-			URL:          message.URL,
-			FileLength:   message.FileLength,
-			CreatedAt:    message.CreatedAt.Format(time.RFC3339),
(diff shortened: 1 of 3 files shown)


src/usecase/chat.go
```diff
@@ -165,19 +165,20 @@ func (service serviceChat) GetChatMessages(ctx context.Context, request domainCh
...
-			UpdatedAt:    message.UpdatedAt.Format(time.RFC3339),
+			ID:               message.ID,
+			ChatJID:          message.ChatJID,
+			SenderJID:        message.Sender,
+			Content:          message.Content,
+			Timestamp:        message.Timestamp.Format(time.RFC3339),
+			IsFromMe:         message.IsFromMe,
+			MediaType:        message.MediaType,
+			CallMetadata:     message.CallMetadata,
+			ReferralMetadata: message.ReferralMetadata,
```
aldinokemal, line 176: 🟠 **Major — rebase without regressing newer response fields.** This initializer is based on a merge base that is 111 commits behind current `main`. The current implementation now also resolves `SenderDisplayName` and maps `Reactions`; GitHub already marks this PR non-mergeable. When resolving the conflict, add `ReferralMetadata: message.ReferralMetadata` to the current initializer rather than taking this older block wholesale, otherwise `/chat/:jid/messages` will silently lose those newer fields.

aldinokemal (commented):
Reviewed the storage-to-API mapping, JSON compatibility, OpenAPI change, current `main` evolution, prior discussion, and available CI status.

The referral metadata mapping itself is additive and preserves `omitempty`, but this branch is now 111 commits behind `main` and GitHub reports it as non-mergeable. The current chat response mapping has since added sender display-name resolution and reaction payloads, so the conflict resolution must preserve those paths while adding `ReferralMetadata`.

No pull-request workflow runs were available for this head SHA, and the PR description says tests were not run. After rebasing, please run `gofmt`, `go test ./...`, and an API-level test covering both a message with valid referral metadata and a normal message where the field must remain omitted.

PR #646, opened 2026-04-24
Show phone number when sending a contact in messages

### LINKED ISSUE #645: Show phone number when sending a contact in messages
## Current behavior
When a contact is sent in a message, only the contact name is shown. The phone number is not included in the shared contact card.

## Expected behavior
When sending a contact, the shared contact should include both:
- contact name
- phone number

## Why this matters
Showing the phone number makes the shared contact easier to verify, especially when multiple contacts have similar names.

## Suggested improvement
Update the contact-send flow so the payload and message preview include the contact phone number, not just the name stored in the database.

juliomuhlbauer:
This PR updates contact message text so the phone number is included alongside the contact name.

Changes:
- outgoing contact messages now store name + phone in the saved content
- contact message text extraction now includes the phone number from the vCard
- added unit tests for proto and event contact text formatting

Closes #645


src/validations/send_validation.go
```diff
@@ -39,15 +40,18 @@ func validateDuration(dur *int) error {
...
 func validatePhoneNumber(phone string) error {
-	if phone == "" {
+	phoneNumber := strings.TrimSpace(phone)
+	if phoneNumber == "" {
 		return pkgError.ValidationError("phone number cannot be empty")
 	}
 
 	// Remove + prefix if present for validation
-	phoneNumber := phone
 	if len(phoneNumber) > 0 && phoneNumber[0] == '+' {
```
aldinokemal, line 49: One edge case still slips through here: `++` passes validation because we only remove the first `+`. Later `CleanPhoneForWhatsApp` removes all plus signs, so `SendContact` can build `waid=:+` and store content without a phone. Can we validate against the same cleaned value used in `SendContact` (or reject when no digits remain after removing `+`) and add a test for `++`?

src/pkg/utils/whatsapp.go
```diff
@@ -57,9 +57,38 @@ func resolveKnownDocumentExtension(mimeType string) (string, bool) {
...
+		}
+		current.WriteString(line)
+	}
+	if current.Len() > 0 {
+		lines = append(lines, current.String())
+	}
+
+	for _, line := range lines {
 		line = strings.TrimSpace(line)
 		if strings.HasPrefix(strings.ToUpper(line), "TEL") {
```
aldinokemal, line 93: Small vCard compatibility case: grouped properties are valid, for example `item1.TEL;type=CELL:+15550100`. Since this only checks lines starting with `TEL`, those cards return no phone and the new contact summary loses the number. Can we match `TEL` after an optional `<group>.` prefix and cover it in `TestExtractPhoneFromVCard`?

aldinokemal (commented):
Left a couple of inline notes from manual review.

PR #648, opened 2026-04-24
feat(chatstorage): persist WhatsApp ContextInfo and expose in /chat/:jid/messages

lucaslorentz:
## Problem

Reply context (`whatsmeow.ContextInfo.StanzaID`) is available in real-time webhook payloads (`replied_to_id`, `quoted_body`) but is **not persisted**. Consumers that process messages asynchronously — notably anything reacting to a prior message — have no way to know that message was a reply, because `GET /chat/:jid/messages` doesn't expose that information.

Concrete scenario: a bot listens to `message.reaction` webhooks, then calls `GET /chat/:jid/messages` to resolve the reacted message. If the reacted message itself was a reply, the bot can't reconstruct that link — the field simply doesn't exist in the response.

## Approach

Add a single `context_metadata TEXT DEFAULT ''` column to the `messages` table (migration 17) storing a JSON blob derived from `ContextInfo`. Shape is open-ended so additional fields (mentions, forward info) can land later without another migration.

Pattern matches `call_metadata` (Migration 15) and `referral_metadata` (Migration 16) — both JSON blobs stored alongside the core fields.

## Current scope

Only `replied_to_id` is captured for now:

\`\`\`json
{"replied_to_id":"AC625C65EA7143E71D784176C4A90A93"}
\`\`\`

Enough to unblock reply resolution. `quoted_body`, `mentioned_jids`, `is_forwarded`, etc. can follow without schema changes.

## Changes

- **Migration 17** — \`ALTER TABLE messages ADD COLUMN context_metadata TEXT DEFAULT ''\`
- **\`domains/chatstorage.Message\`** — new \`ContextMetadata string\` field with \`db:\"context_metadata\"\` tag
- **\`infrastructure/chatstorage/sqlite_repository.go\`** — INSERT/UPDATE (both \`StoreMessage\` and \`StoreMessagesBatch\`), SELECTs, and \`scanMessage\` include the new column
- **\`CreateMessage\`** (inbound ingestion) — extracts \`ContextInfo.StanzaID\` via \`utils.ExtractContextInfo\` and marshals into the blob
- **\`CreateSentMessage\`** (outgoing bot messages) — same extraction, so bot-authored replies are also recoverable
- **\`domains/chat.MessageInfo\`** — new \`ContextMetadata string \`json:\"context_metadata,omitempty\"\`\` field
- **\`usecase/chat.go\`** — mapper copies the field through

## Response shape

For messages with context:

\`\`\`json
{
  \"id\": \"AC4BFC5C3AD9CFB3797F8E9F8D2968A0\",
  \"content\": \"got it\",
  \"context_metadata\": \"{\\\"replied_to_id\\\":\\\"AC625C65EA7143E71D784176C4A90A93\\\"}\",
  ...
}
\`\`\`

Consumers \`json.Unmarshal\` when present. Messages without context omit the field entirely (\`omitempty\`), so existing responses look unchanged.

## Webhook payloads unchanged

\`replied_to_id\` and \`quoted_body\` still show up as flat top-level fields in \`message\`-event webhook payloads (\`infrastructure/whatsapp/event_message.go\`). Existing webhook consumers are not affected.

## Backward compatibility

- Additive migration (\`ADD COLUMN ... DEFAULT ''\`).
- \`omitempty\` keeps old responses byte-identical for messages without context.
- No renames, no type changes, no behavior change for non-reply messages.

## Validation

Fresh build against \`main\` HEAD on darwin/arm64, deployed over a live instance with schema previously at version 14. Migrations 15–17 ran cleanly. Sent a WhatsApp reply; verified:

- \`chatstorage.db\` row has \`context_metadata = '{\"replied_to_id\":\"...\"}'\` on the reply, empty on the quoted message.
- \`GET /chat/:jid/messages\` response includes the \`context_metadata\` string.
- Pre-migration messages show empty \`context_metadata\` → omitted from response, no visible change.

## Related gap (separate)

\`referral_metadata\` is stored but not mapped into \`MessageInfo\` (no struct field, no mapper copy). Kept out of scope here; can follow up in a tiny separate PR if interested.

### DIFF
--- a/src/infrastructure/chatstorage/sqlite_repository.go
+++ b/src/infrastructure/chatstorage/sqlite_repository.go
@@ -27,6 +27,29 @@ func NewStorageRepository(db *sql.DB) domainChatStorage.IChatStorageRepository {
 	return &SQLiteRepository{db: db}
 }
 
+// buildContextMetadata marshals whatsmeow ContextInfo-derived fields into a
+// JSON blob for the messages.context_metadata column. Returns "" when the
+// message has no meaningful context. Shape is open so future fields
+// (mentioned_jids, is_forwarded, ...) can be added without a migration.
+func buildContextMetadata(ci *waE2E.ContextInfo) string {
+	if ci == nil {
+		return ""
+	}
+	meta := map[string]any{}
+	if stanzaID := ci.GetStanzaID(); stanzaID != "" {
+		meta["replied_to_id"] = stanzaID
+	}
+	if len(meta) == 0 {
+		return ""
+	}
+	jsonBytes, err := json.Marshal(meta)
+	if err != nil {
+		logrus.WithError(err).Warn("failed to marshal context metadata")
+		return ""
+	}
+	return string(jsonBytes)
+}
+
 // StoreChat creates or updates a chat
 func (r *SQLiteRepository) StoreChat(chat *domainChatStorage.Chat) error {
 	now := time.Now()
@@ -89,7 +112,7 @@ func (r *SQLiteRepository) GetMessageByID(id string) (*domainChatStorage.Message
 	query := `
 		SELECT id, chat_jid, device_id, sender, content, timestamp, is_from_me,
 			media_type, call_metadata, filename, url, media_key, file_sha256,
-			file_enc_sha256, file_length, referral_metadata, created_at, updated_at
+			file_enc_sha256, file_length, referral_metadata, context_metadata, created_at, updated_at
 		FROM messages
 		WHERE id = ?
 		LIMIT 1
@@ -317,11 +340,11 @@ func (r *SQLiteRepository) StoreMessage(message *domainChatStorage.Message) erro
 	result, err := r.db.Exec(`
 		UPDATE messages SET sender = ?, content = ?, timestamp = ?, is_from_me = ?,
 			media_type = ?, call_metadata = ?, filename = ?, url = ?, media_key = ?, file_sha256 = ?,
-			file_enc_sha256 = ?, file_length = ?, referral_metadata = ?, updated_at = ?
+			file_enc_sha256 = ?, file_length = ?, referral_metadata = ?, context_metadata = ?, updated_at = ?
 		WHERE id = ? AND chat_jid = ? AND device_id = ?
 	`, message.Sender, message.Content, message.Timestamp, message.IsFromMe,
 		message.MediaType, message.CallMetadata, message.Filename, message.URL, message.MediaKey, message.FileSHA256,
-		message.FileEncSHA256, message.FileLength, message.ReferralMetadata, message.UpdatedAt,
+		message.FileEncSHA256, message.FileLength, message.ReferralMetadata, message.ContextMetadata, message.UpdatedAt,
 		message.ID, message.ChatJID, message.DeviceID)
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


src/infrastructure/chatstorage/sqlite_repository.go
```diff
@@ -391,7 +414,7 @@ func (r *SQLiteRepository) StoreMessagesBatch(messages []*domainChatStorage.Mess
 		result, err := updateStmt.Exec(
 			message.Sender, message.Content, message.Timestamp, message.IsFromMe,
 			message.MediaType, message.CallMetadata, message.Filename, message.URL, message.MediaKey, message.FileSHA256,
-			message.FileEncSHA256, message.FileLength, message.ReferralMetadata, message.UpdatedAt,
+			message.FileEncSHA256, message.FileLength, message.ReferralMetadata, message.ContextMetadata, message.UpdatedAt,
```
aldinokemal, line 417: 🟠 **Major — history-synced replies still lose this metadata.** `StoreMessagesBatch` now writes `message.ContextMetadata`, but `processConversationMessages` in `infrastructure/whatsapp/history_sync.go` constructs each `domainChatStorage.Message` without extracting `ContextInfo`, so this value is always empty for INITIAL_BOOTSTRAP/RECENT syncs. Populate the field from `msg.GetMessage()` using the same extraction logic (preferably a shared helper), and add a history-sync regression test with a quoted message.

src/infrastructure/chatstorage/sqlite_repository.go
```diff
@@ -2103,5 +2142,10 @@ func (r *SQLiteRepository) getMigrations() []string {
 
 		// Migration 29: Fetch due Chatwoot retry jobs in stable order
 		`CREATE INDEX IF NOT EXISTS idx_chatwoot_forward_queue_due ON chatwoot_forward_queue(next_attempt_at, id)`,
+
+		// Migration 30: JSON metadata for whatsmeow ContextInfo (reply refs,
+		// mentions, forward info). Open-ended so future fields don't need a new
+		// migration — consumers JSON-parse on read.
+		`ALTER TABLE messages ADD COLUMN context_metadata TEXT DEFAULT ''`,
```
aldinokemal, line 2149: 🔴 **Rebase blocker — migration number 30 is already occupied on current `main` by `direct_path`; current main now ends at migration 43.** This schema change must be appended as the next migration after rebasing (currently 44), not inserted at the old branch index. Also re-check every SELECT/INSERT/scan argument against the newer `direct_path` and device/webhook columns so column ordering is not lost during conflict resolution.

aldinokemal (commented):
Reviewed the live inbound path, outbound protobuf path, batch/history-sync path, SQL column ordering, API mapping, migration evolution, previous discussion, and available CI status.

The live `CreateMessage` and outbound `StoreSentMessageWithContext` paths correctly extract `ContextInfo.StanzaID`, but the implementation does not yet cover history-synced replies. That means an initial/recent history sync still stores empty `context_metadata`, even though the PR extends `StoreMessagesBatch` to persist it. This is a functional gap for exactly the asynchronous consumers described in the PR.

The branch has also diverged substantially from `main` (main is 54 commits ahead), GitHub reports it as non-mergeable, and no pull-request workflow runs were available for the head SHA. After rebasing, please preserve the newer message fields/device-scoped logic, append the schema change after the current migration sequence, update OpenAPI for the new response field, and add tests covering live inbound, outbound reply, history-sync reply, non-reply omission, and migration from the latest released schema.

PR #654, opened 2026-04-27
Persist incoming contact messages in chatstorage

### LINKED ISSUE #651: Persist incoming ContactMessage events in chatstorage
### Problem
Incoming WhatsApp contact messages (`ContactMessage`) are not persisted to the local chatstorage database. The current storage path only extracts text, image, video, document, etc., and returns early when both content and media type are empty.

### Current behavior
- `handleMessage` forwards every incoming message to `CreateMessage(...)`.
- `CreateMessage(...)` uses `ExtractMessageTextFromProto(...)` and `ExtractMediaInfo(...)`.
- `ContactMessage` is only used for webhook payloads / display text, but not for database persistence.
- As a result, a received contact card is dropped from the local history.

### Expected behavior
- A received contact card should be stored in the `messages` table, similar to other message types.
- The stored row should preserve at least the contact display name and a marker that it is a contact-type message.

### Suggested acceptance criteria
- Incoming `ContactMessage` events are saved to chat history.
- They appear in message lists and exports.
- Existing behavior for text, image, video, and document messages remains unchanged.

### Notes
This is different from sending a contact card via `/send/contact`, which already stores the sent message as a normal message row in chatstorage.

juliomuhlbauer:
Closes #651

This PR teaches incoming `ContactMessage` events to serialize into chat history instead of being dropped when the message has no text or media. It also adds coverage for `ContactMessage` parsing so the phone number embedded in the vCard is preserved in stored content.


src/pkg/utils/whatsapp.go
```diff
@@ -56,6 +56,18 @@ func resolveKnownDocumentExtension(mimeType string) (string, bool) {
...
+func extractPhoneFromVCard(vcard string) string {
+	for _, line := range strings.Split(vcard, "\n") {
+		line = strings.TrimSpace(line)
+		if strings.HasPrefix(strings.ToUpper(line), "TEL") {
+			if idx := strings.LastIndex(line, ":"); idx >= 0 {
+				return strings.TrimSpace(line[idx+1:])
+			}
+		}
+	}
+	return ""
```
aldinokemal, line 68: this is duplicated in `infrastructure/whatsapp/webhook_forward.go`, maybe export it here and reuse so they don't drift

src/pkg/utils/whatsapp.go
```diff
@@ -138,6 +150,43 @@ func ExtractMessageTextFromProto(msg *waE2E.Message) string {
...
+				case name != "" && phone != "":
+					return fmt.Sprintf("Contacts: %s (%s)", name, phone)
+				case name != "":
+					return "Contacts: " + name
+				case phone != "":
+					return "Contacts: " + phone
+				}
+			}
+		}
+		return "Contacts shared"
```
aldinokemal, line 187: if there's more than one contact, only the first one is kept and the rest are silently dropped. could add `+N more` or include the count so it's not lost

src/pkg/utils/whatsapp.go
juliomuhlbauer: Ajustado: extraí `ExtractPhoneFromVCard` e `FormatContactText`/`FormatContactsText` para `pkg/utils` e passei o webhook a reutilizar esses helpers, removendo a cópia local para evitar drift.

src/pkg/utils/whatsapp.go
juliomuhlbauer: Ajustado: o fluxo agora preserva o primeiro contato e anexa `+N more` quando há múltiplos contatos; quando não há nome/telefone úteis, volta para `Contacts shared`.

aldinokemal:
nice fix, the approach is clean, just letting the extractor return non-empty so the existing skip-gate in `sqlite_repository.go` lets it through.

a couple of small things worth a look:

1. `extractPhoneFromVCard` already exists in `infrastructure/whatsapp/webhook_forward.go` (line ~422), and the same name/phone formatting switch is in `structuredContactsArraySummary` there too. having two copies will drift over time, and these are the two places users actually compare (stored history vs webhook payload). probably worth exporting one `ExtractPhoneFromVCard` from `pkg/utils` and deleting the duplicate.

2. `ExtractMessageTextFromEvent` (same file, line ~254) still only returns `DisplayName` for contact messages and has no `ContactsArrayMessage` branch. after this PR the two extractors disagree on the same event — easy to align while you're in there.

minor: tests cover `ContactMessage` but not `ContactsArrayMessage` or the empty-name/empty-phone fallback. the vCard split on `\n` is fine because `TrimSpace` eats the `\r`, but a one-line note would help future readers.

LGTM otherwise 👍

aldinokemal (approved):

PR #659, opened 2026-05-03
Forward WhatsApp label appstate events to webhooks

zayedadel:
## Summary
* Forward WhatsApp label edit and chat label association appstate updates to configured webhooks.
* Reuse existing webhook forwarding, event filtering, and LID normalization behavior.
* Keep payloads small and predictable with label metadata, label state, chat_id, and chat_lid when applicable.

## Review Notes
* Addressed CodeRabbit's request by preserving chat_id with the raw value when a label association target cannot be parsed as a JID.
* Avoids starting a webhook goroutine for unrelated appstate updates.

## Tests
* go test ./...


zayedadel:
Thanks for the review. I updated the PR to keep the label webhook handling smaller and aligned with the existing event forwarding conventions.

The CodeRabbit finding is addressed by preserving `chat_id` with the raw appstate value when `types.ParseJID` fails, and there is now a regression test for that case.

@coderabbitai review

src/infrastructure/whatsapp/event_label.go
```diff
@@ -0,0 +1,123 @@
...
+	eventTypeLabelEdit        = "label.edit"
+	eventTypeLabelAssociation = "label.association"
+)
+
+func isLabelAppState(evt *events.AppState) bool {
+	if evt == nil || len(evt.Index) == 0 {
+		return false
+	}
+
+	return evt.Index[0] == appstate.IndexLabelEdit || evt.Index[0] == appstate.IndexLabelAssociationChat
```
aldinokemal, line 26: **Suggestion (Maintainability):** `appstate.IndexLabelAssociationMessage` (`label_message`) is intentionally excluded — only chat-level label associations are forwarded. Worth a 1-line comment so the next reader doesn't think it was overlooked.

```go
// Note: IndexLabelAssociationMessage (per-message label) is intentionally excluded.
// Only chat-level associations and label edits are forwarded.
return evt.Index[0] == appstate.IndexLabelEdit || evt.Index[0] == appstate.IndexLabelAssociationChat
```

src/infrastructure/whatsapp/event_label.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+		addLabelChatFields(ctx, client, payload, evt.Index[2])
+
+		return eventTypeLabelAssociation, payload
+	default:
+		return "", nil
+	}
+}
+
+func addLabelEditActionFields(payload map[string]any, action *waSyncAction.LabelEditAction) {
```
aldinokemal, line 79: **Suggestion (Completeness):** `LabelEditAction` has three additional fields not included in the payload:
- `IsActive *bool` (proto field 6)
- `IsImmutable *bool` (proto field 8)
- `MuteEndTimeMS *int64` (proto field 9)

If "small and predictable" is the intent (per PR description), fine — but `IsActive` is arguably as informative as `Deleted` and consumers might expect it. Either include them or add a comment explaining why they're omitted.

src/infrastructure/whatsapp/event_label.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		payload["chat_id"] = rawJID
+
+		return
+	}
+
+	chatJID := jid.ToNonAD()
+	if chatJID.Server == "lid" {
+		payload["chat_lid"] = chatJID.String()
+		if client != nil && client.Store != nil && client.Store.LIDs != nil {
```
aldinokemal: **Suggestion (Code clarity):** This nil-check guard is redundant — `NormalizeJIDFromLID` already nil-checks `client`, `client.Store`, and `client.Store.LIDs` internally and falls back gracefully. You can simplify to:

```go
if chatJID.Server == "lid" {
    payload["chat_lid"] = chatJID.String()
    chatJID = NormalizeJIDFromLID(ctx, chatJID, client).ToNonAD()
}
```

**Side note:** When `NormalizeJIDFromLID` fails to resolve, both `chat_id` and `chat_lid` end up holding the same LID string — consumers can't distinguish "resolved" from "not resolved". Worth documenting in the webhook payload spec.

src/infrastructure/whatsapp/event_label.go
```diff
@@ -0,0 +1,123 @@
...
+			chatJID = NormalizeJIDFromLID(ctx, chatJID, client).ToNonAD()
+		}
+	}
+
+	payload["chat_id"] = chatJID.String()
+}
+
+func labelAppStateTimestamp(evt *events.AppState) string {
+	if evt != nil && evt.SyncActionValue != nil && evt.Timestamp != nil && evt.GetTimestamp() > 0 {
```
aldinokemal: **Suggestion (Readability):** Over-defensive check. `GetTimestamp()` already nil-checks `Timestamp` internally and returns 0 if nil, and `evt.SyncActionValue != nil` is already guaranteed by the guard in `buildLabelAppStatePayload` (which is the only caller path that reaches here). Could simplify to:

```go
func labelAppStateTimestamp(evt *events.AppState) string {
    if ts := evt.GetTimestamp(); ts > 0 {
        return time.UnixMilli(ts).UTC().Format(time.RFC3339)
    }
    return time.Now().UTC().Format(time.RFC3339)
}
```

src/infrastructure/whatsapp/event_label_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+	if len(actual) != len(expected) {
+		t.Fatalf("expected payload length %d, got %d: %#v", len(expected), len(actual), actual)
+	}
+	for key, expectedValue := range expected {
+		actualValue, ok := actual[key]
+		if !ok {
+			t.Fatalf("expected payload key %q", key)
+		}
+		if actualValue != expectedValue {
```
aldinokemal: **Suggestion (Test robustness):** `actualValue != expectedValue` works for the current primitives (string, bool, int32) but will silently pass mismatched maps/slices if the payload structure evolves later. Future-proof with `reflect.DeepEqual`:

```go
if !reflect.DeepEqual(actualValue, expectedValue) {
    t.Fatalf("expected payload[%q] to be %#v, got %#v", key, expectedValue, actualValue)
}
```

(requires `import "reflect"`)

src/infrastructure/whatsapp/event_label_test.go
```diff
@@ -0,0 +1,146 @@
...
+import (
+	"context"
+	"testing"
+
+	"go.mau.fi/whatsmeow/proto/waSyncAction"
+	"go.mau.fi/whatsmeow/types/events"
+	"google.golang.org/protobuf/proto"
+)
+
+func TestBuildLabelAppStatePayload(t *testing.T) {
```
aldinokemal, line 14: **Suggestion (Test coverage):** Defensive guards in `buildLabelAppStatePayload` and `labelAppStateTimestamp` are not exercised. Consider adding cases for:

- `evt = nil`
- `evt.SyncActionValue = nil`
- `Index` shorter than required (`len < 2` for `label_edit`, `len < 3` for `label_jid`)
- `LabelEditAction == nil` with index `label_edit`
- `LabelAssociationAction == nil` with index `label_jid`

Also consider a test for `labelAppStateTimestamp` covering: missing timestamp (fallback to `time.Now`), zero timestamp, and a valid millisecond timestamp.

aldinokemal (commented):
## Code Review Summary

**Verdict: Approve with minor suggestions** ✅

Solid implementation, consistent with existing event patterns (`event_group.go`, `event_receipt.go`). Defensively nil-checks, normalizes LIDs, preserves raw `chat_id` on parse failure (per CodeRabbit feedback). No critical issues.

### What looks good
- Re-uses `forwardPayloadToConfiguredWebhooks` — gets webhook whitelisting and partial-failure handling for free
- Conditional field inclusion via proto optional getters keeps payload shape clean
- Early `len(config.WhatsappWebhook) > 0 && isLabelAppState(evt)` short-circuit avoids spawning goroutines for unrelated AppState events
- Goroutine pattern with 30s `context.WithTimeout` matches existing handlers
- `chat_lid` exposed alongside resolved `chat_id` mirrors `event_message.go::buildFromFields`
- No race conditions, no goroutine leaks

### Security & Performance
- No injection vectors. Webhook URLs come from validated config.
- Label edits are infrequent — no rate-limit concerns.
- No N+1 queries; `NormalizeJIDFromLID` is a single store lookup.

See inline comments below for minor suggestions. Items #4 (test robustness) and #5 (missing edge case tests) are the highest-value follow-ups.

src/infrastructure/whatsapp/event_label.go
zayedadel: done

src/infrastructure/whatsapp/event_label.go
```diff
@@ -0,0 +1,143 @@
...
+		Event:    EventTypeLabelAssociation,
+		DeviceID: deviceID,
+		Payload:  payload,
+	}
+}
+
+func addLabelTargetJIDFields(ctx context.Context, client *whatsmeow.Client, payload map[string]any, rawJID string) {
+	jid, err := types.ParseJID(rawJID)
+	if err != nil {
+		return
```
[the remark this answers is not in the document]
zayedadel: Addressed in commit facf8ca. `addLabelChatFields` now sets `payload["chat_id"] = rawJID` in the error path before returning, so consumers always receive `chat_id` regardless of whether `types.ParseJID` succeeds. A regression test case (`label association with unparsable chat id`) covers this path.

src/infrastructure/whatsapp/event_label.go
zayedadel: Added in commit facf8ca. `isLabelAppState` now has a two-line comment above the return statement explaining that `IndexLabelAssociationMessage` (per-message label) is intentionally excluded and only chat-level associations and label edits are forwarded.

src/infrastructure/whatsapp/event_label.go
zayedadel: Added in commit facf8ca. `addLabelEditActionFields` now includes nil-guarded entries for `is_active` (`IsActive`), `is_immutable` (`IsImmutable`), and `mute_end_time_ms` (`MuteEndTimeMS`), following the same pattern as the existing fields. The `label edit` test case is updated to set and assert all three.

src/infrastructure/whatsapp/event_label.go
zayedadel: Simplified in commit facf8ca. The guard is now just `if client != nil` — `NormalizeJIDFromLID` handles `client.Store` and `client.Store.LIDs` nil-checks internally and falls back to the original JID.

src/infrastructure/whatsapp/event_label.go
zayedadel: Simplified in commit facf8ca. `labelAppStateTimestamp` now uses `if ts := evt.GetTimestamp(); ts > 0` directly, dropping the outer nil guards on `evt`, `evt.SyncActionValue`, and `evt.Timestamp`.

src/infrastructure/whatsapp/event_label_test.go
zayedadel: Updated in commit facf8ca. `assertPayloadEqual` now uses `reflect.DeepEqual` instead of `!=`, with the `"reflect"` import added.

src/infrastructure/whatsapp/event_label_test.go
zayedadel: Added in commit facf8ca. `TestBuildLabelAppStatePayload` now includes six additional defensive cases: nil `evt`, nil `SyncActionValue`, short index for `label_edit`, nil `LabelEditAction`, short index for `label_jid`, and nil `LabelAssociationAction`. A new `TestLabelAppStateTimestamp` covers: empty event (fallback to now), zero timestamp (fallback to now), and a valid millisecond timestamp formatted to RFC3339.

zayedadel:
@aldinokemal thanks for ur active maintaince of the project , and not ignoring pull request  

if no more changes are requiered for this pr i intend to address  webhook/event translation for poll votes behavior 

Right now:

poll creation is supported and sent correctly
when a user votes, WhatsApp emits a PollUpdateMessage
the repo currently forwards that as a generic message event with no body
downstream apps only see an empty inbound message, so they cannot know which option was chosen

do u have any guidance or conventions/opinions i have to keep in mind while i tackle this ?

PR #662, opened 2026-05-04
Separate sender display name

juliomuhlbauer:
What changed
- Added `sender_display_name` to chat message payloads.
- Resolved sender display names from the WhatsApp contact store when available, with a fallback to the sender JID.
- Updated the chat messages UI to render the new field instead of the nonexistent `push_name` property.

Why
- The UI was relying on a field that was never sent by the backend, which made sender labels fall back inconsistently and could surface the wrong contact-facing name.

Impact
- Message history now shows a dedicated sender display name without reusing chat names.
- Existing behavior for messages still falls back safely to the sender JID.

Validation
- `cd src && /opt/homebrew/bin/go test ./...`


aldinokemal:
Thank you for the contribution. This PR is now superseded by #782, which implements the maintained dynamic sender display-name design across REST chat history, reactions, and all sender-bearing webhooks while preserving the existing compatibility fields. I am closing this PR to keep review focused on the maintained implementation; the contributor branch is left intact.

PR #666, opened 2026-05-05
Persist reply metadata in message storage

juliomuhlbauer:
Stores replied_to_id and quoted_body for incoming and sent messages, and adds SQLite coverage for both reply flows.


src/infrastructure/chatstorage/sqlite_repository.go
```diff
@@ -507,11 +510,25 @@ func (r *SQLiteRepository) scanMessage(scanner interface{ Scan(...any) error })
...
 
+func extractReplyContext(msg *waE2E.Message) (repliedToID, quotedBody string) {
+	ci := utils.ExtractContextInfo(utils.UnwrapMessage(msg))
+	if ci == nil {
+		return "", ""
+	}
+
+	repliedToID = ci.GetStanzaID()
+	if quoted := ci.GetQuotedMessage(); quoted != nil {
+		quotedBody = quoted.GetConversation()
```
aldinokemal: This only grabs `Conversation`, so `quoted_body` stays empty when the quoted message is extended text or a media caption. Can we use the existing text extraction helper here, or add the same fallbacks for extended text + image/video/document captions? Please add a small test for one of those too, so this does not silently regress later.

aldinokemal (changes requested):
Need one small fix before merge. The storage/query/migration wiring looks fine, and I verified the real PR diff is only the three chat-storage files. The remaining issue is that `quoted_body` is currently only populated for plain conversation messages, so common quoted message types still come back empty.

juliomuhlbauer:
@aldinokemal let me know if I missed something

src/infrastructure/chatstorage/sqlite_repository.go
```diff
@@ -365,7 +366,7 @@ func (r *SQLiteRepository) StoreMessagesBatch(messages []*domainChatStorage.Mess
 		result, err := updateStmt.Exec(
 			message.Sender, message.Content, message.Timestamp, message.IsFromMe,
 			message.MediaType, message.CallMetadata, message.Filename, message.URL, message.MediaKey, message.FileSHA256,
-			message.FileEncSHA256, message.FileLength, message.ReferralMetadata, message.UpdatedAt,
+			message.FileEncSHA256, message.FileLength, message.ReferralMetadata, message.RepliedToID, message.QuotedBody, message.UpdatedAt,
```
aldinokemal, line 369: Small blocker here: `StoreMessagesBatch` now updates `replied_to_id` and `quoted_body` from the struct, but `src/infrastructure/whatsapp/history_sync.go` builds these batch messages from real WhatsApp history without filling those fields. That means a later history sync can overwrite reply metadata captured by the live/sent paths with empty strings. Please extract the reply context in the history-sync path before `StoreMessagesBatch`, and add a regression test where batch update preserves/populates these fields.

aldinokemal (changes requested):
One real blocker before merge: the live/sent storage paths look good, but the history-sync batch path can erase the new reply metadata. Details inline.

src/infrastructure/chatstorage/sqlite_repository.go
juliomuhlbauer: Commited the changes, let me know if I missed something

aldinokemal:
Thanks for the contribution! Closing this as a duplicate of #648, which persists the same reply context (`replied_to_id`) using the existing `*_metadata` JSON-blob pattern (`call_metadata` / `referral_metadata`), so it can be extended without new migrations — and it already has an approving review. The extra pieces here (`quoted_body`, history-sync coverage) are worth porting into #648 as a follow-up.

PR #673, opened 2026-05-09
Fix chatstorage device ids

juliomuhlbauer:
## What changed
- Normalize storage `device_id` values so chat/message writes and reads use the same JID form.
- Refresh per-device chat storage wrappers when a device gains a live JID/client.
- Add tests covering JID normalization and the storage wrapper behavior.

## Why
The storage layer could receive a raw WhatsApp device JID like `...:11@s.whatsapp.net` or a placeholder device ID, while queries elsewhere expect the normalized JID form. That mismatch made the storage look like it was not updating.

## Validation
- `go test ./...`


juliomuhlbauer:
@aldinokemal This is critical, the webhook and storage are not getting outbound messages.
This problem seems to be started in v8.5.0.
I don't know if this fixes it.

aldinokemal (changes requested):
I think this needs one small fix before we merge.

The normalization looks mostly right, but there is still a stale wrapper path for first-time paired devices. In `EnsureClient`, `refreshChatStorage(inst)` runs right after `SetClient` / `UpdateStateFromClient`, but at that point a QR / pair-code device may still only have the placeholder device ID. Later, when the real JID is populated in `handleConnectionEvents`, the chat storage wrapper is not refreshed again.

That can leave wrapper-based writes scoped to the placeholder ID while reads and other storage paths expect the normalized JID, so the same bug can still show up after pairing.

Can we refresh the per-device chat storage after the JID is updated in `handleConnectionEvents`, and add a small regression test for placeholder ID -> real JID storage scoping? After that I'm good with this.

juliomuhlbauer:
@aldinokemal let me know if I missed somehting

src/infrastructure/whatsapp/device_manager.go
```diff
@@ -501,6 +514,7 @@ func (m *DeviceManager) EnsureClient(ctx context.Context, deviceID string) (*Dev
 
 	inst.SetClient(client)
 	inst.UpdateStateFromClient()
+	m.refreshChatStorage(inst)
```
aldinokemal, line 517: Almost there, but this refresh only covers the setup/reuse path in `EnsureClient`. For a brand-new QR / pair-code device, the instance can still be on the placeholder ID here, then the real WhatsApp JID shows up later when `handleConnectionEvents` calls `UpdateStateFromClient()`.

Can you refresh the device chat storage again right after that connection-event JID update, and add a small placeholder ID -> real JID test? That should keep wrapper-based writes from staying stuck under the placeholder device ID.

juliomuhlbauer: @aldinokemal Let me know if everything is ok now
We really need to solve this for our app

juliomuhlbauer:
@aldinokemal something missing from my side? we can't update gowa to the latest version because of this

src/infrastructure/whatsapp/event_handler.go
```diff
@@ -183,6 +183,9 @@ func handleConnectionEvents(_ context.Context, client *whatsmeow.Client, instanc
 	}
 	if instance != nil {
 		instance.UpdateStateFromClient()
+		if instanceManager := GetDeviceManager(); instanceManager != nil {
```
aldinokemal, line 186: This is the right lifecycle point for the original placeholder→live-JID bug, and the added regression test addresses my earlier thread. However, this branch predates the current-main `handleConnectionEvents`, which now also persists `ADJID`, updates/invalidate the per-device Chatwoot config JID, and triggers Chatwoot auto-sync. Please rebase and keep this refresh immediately after `UpdateStateFromClient()` without replacing/dropping those current-main behaviors.

src/usecase/send.go
```diff
@@ -62,14 +67,14 @@ func (service serviceSend) wrapSendMessage(ctx context.Context, client *whatsmeo
...
-		senderJID = client.Store.ID.String()
+	if client != nil && client.Store != nil && client.Store.ID != nil {
+		senderJID = client.Store.ID.ToNonAD().String()
 	}
 
 	// Store message asynchronously with timeout.
 	// Preserve device context (for device_id scoping) but detach from request cancellation.
 	go func() {
-		storeCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 2*time.Second)
+		storeCtx, cancel := buildSentMessageStoreContext(ctx, 2*time.Second)
```
aldinokemal, line 77: Please do not carry this 2s budget forward when rebasing. Current main intentionally raised sent-message persistence to 15s because history-sync batches / SQLite writer contention can outlive 2s and otherwise leave successfully sent messages missing from chat storage. Current main also removed the old `SendMessageWithReachoutRetry` path in favor of direct `client.SendMessage`; preserve both current-main decisions while keeping the `ToNonAD()` sender normalization.

aldinokemal (changes requested):
The original owner blocker is fixed now: the connection-event path refreshes the wrapper after the live JID becomes available, and the placeholder→real-JID regression test covers that transition, so I resolved that old thread.

The PR is not merge-safe against current `main`, though. Its base is still `fb5a0613…` while current main is `b2a57a20…`, GitHub reports merge conflicts, and several assumptions in this branch have since changed. Please rebase onto current main and re-apply the storage-ID fix without dropping the newer DeviceManager/Chatwoot/send-path behavior.

Two concrete current-main regressions must be avoided during that rebase: (1) current `handleConnectionEvents` now persists JID/AD-JID, synchronizes the per-device Chatwoot config JID/registry, and triggers Chatwoot auto-sync; the refresh belongs immediately after `UpdateStateFromClient`, but those newer behaviors must remain intact; (2) current `wrapSendMessage` deliberately uses direct `client.SendMessage` and a 15s detached chat-storage budget because history-sync/SQLite writer contention can exceed the old 2s budget. This branch still carries the old reachout-retry path and 2s storage timeout.

Also update the new `recordingChatStorage` test double after rebasing: current `IChatStorageRepository` now includes device-scoped message lookups, poll definition methods, Chatwoot correlation/config/queue methods, `GetDeviceRecordByJID`, and webhook-config methods, so this test helper will not satisfy the current interface as written.

The core JID idea itself is still sound with current pinned `go.mau.fi/whatsmeow@0fadda796019`: using `ToNonAD()` for the chat-storage partition/sender identity matches current `DeviceInstance.JID()` semantics. No protocol blocker beyond the stale send path was found.

PR #701, opened 2026-05-29
🐛 fix: detect SecretEncrypted message edits for webhook, storage and UI

### LINKED ISSUE #700: 🐞 Bug: Edited messages are emitted as message instead of message.edited
### What happened?

Description

There seems to be a bug related to WhatsApp edited messages events.

When a message is edited in WhatsApp, the event emitted by the library is still message instead of message.edited.

Because of this, the payload received does not contain the expected body field, making it impossible to properly handle edited messages.

Expected Behavior

When a user edits a message, the library should emit:

{
  "event": "message.edited"
}

And the payload should contain the updated message body.

Current Behavior

Currently, edited messages are emitted as:

{
  "event": "message"
}

However, the payload does not contain the body field.

Possible Impact

Applications relying on message events cannot distinguish between:

new messages
edited messages

Additionally, edited messages become unusable because the updated content is missing.

Steps to Reproduce
Send a message
Edit the message in WhatsApp
Observe the emitted websocket/event payload
The event is emitted as message instead of message.edited
The payload has no body
Environment
go-whatsapp-web-multidevice version: latest
WhatsApp Web Multi Device enabled
OS: Linux

Thanks for the project.

### Reproduction Steps

```bash
Send a message to any chat
Edit the same message in WhatsApp
Observe the emitted event/websocket payload
The library emits the event as message
The payload does not contain the edited body
Expected event should be message.edited
```

### Relevant log output

```bash
 whatsapp_go-1  ..... secretEncryptedMessage:{targetMessageKey:{remoteJID:...secretEncType:MESSAGE_EDIT} 
whatsapp_go-1 | time="...." level=debug msg="Skipping message {HIDE} - no content or media"
```

### Screenshots or Additional Context

_No response_

### Application Version

8.6.0

### Verification

- [x] I have checked for existing closed issues and discussions.

emtudo:
- Add message_edit helpers to decrypt SecretEncryptedMessage(MESSAGE_EDIT) and parse MESSAGE_EDIT protocol payloads
- Route webhook buildEventPayload through ResolveIncomingMessage so edits emit message.edited with body instead of message
- Propagate device context into webhook goroutine for reliable LID/decrypt handling
- Update CreateMessage/storeEditedMessage to resolve encrypted edits before persisting to messages and message_edits
- Broadcast MESSAGE_EDITED over websocket and refresh ChatMessages in the embedded UI
- Add Conversation fallback in BuildEventMessage for edited text extraction
- Add unit tests for edit helpers, webhook payload, and secret-encrypted edit without client

# PR Title

## Context

It resolve: https://github.com/aldinokemal/go-whatsapp-web-multidevice/issues/700

## Test Results

It resolve: https://github.com/aldinokemal/go-whatsapp-web-multidevice/issues/700

### DIFF
--- a/src/infrastructure/chatstorage/sqlite_repository.go
+++ b/src/infrastructure/chatstorage/sqlite_repository.go
@@ -1019,7 +1019,7 @@ func (r *SQLiteRepository) CreateMessage(ctx context.Context, evt *events.Messag
 
 	resolved, err := whatsapp.ResolveIncomingMessage(ctx, client, evt)
 	if err != nil {
-		return nil
+		return fmt.Errorf("failed to resolve incoming message %s: %w", evt.Info.ID, err)
 	}
 
 	if edit := whatsapp.ExtractMessageEdit(resolved); edit != nil {
--- a/src/infrastructure/chatstorage/sqlite_repository.go
+++ b/src/infrastructure/chatstorage/sqlite_repository.go
@@ -1017,7 +1017,12 @@ func (r *SQLiteRepository) CreateMessage(ctx context.Context, evt *events.Messag
 		return fmt.Errorf("failed to get existing chat: %w", err)
 	}
 
-	if editMessage := extractEditedMessage(evt.Message); editMessage != nil {
+	resolved, err := whatsapp.ResolveIncomingMessage(ctx, client, evt)
+	if err != nil {
+		return nil
+	}
+
+	if edit := whatsapp.ExtractMessageEdit(resolved); edit != nil {
 		if existingChat == nil {
 			chat := &domainChatStorage.Chat{
 				DeviceID:        deviceID,
@@ -1030,7 +1035,7 @@ func (r *SQLiteRepository) CreateMessage(ctx context.Context, evt *events.Messag
 			}
 		}
 
-		if err := r.storeEditedMessage(ctx, evt, deviceID, chatJID, sender, editMessage); err != nil {
+		if err := r.storeEditedMessage(ctx, evt, deviceID, chatJID, sender, edit); err != nil {
 			return err
 		}
 		return nil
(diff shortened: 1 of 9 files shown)


src/infrastructure/chatstorage/sqlite_repository.go
```diff
@@ -1017,7 +1017,12 @@ func (r *SQLiteRepository) CreateMessage(ctx context.Context, evt *events.Messag
 		return fmt.Errorf("failed to get existing chat: %w", err)
 	}
 
-	if editMessage := extractEditedMessage(evt.Message); editMessage != nil {
+	resolved, err := whatsapp.ResolveIncomingMessage(ctx, client, evt)
```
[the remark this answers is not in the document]
emtudo: resolved!

src/infrastructure/whatsapp/event_message_handler.go
```diff
@@ -109,28 +112,68 @@ func handleAutoMarkRead(ctx context.Context, evt *events.Message, client *whatsm
...
+			deviceID = inst.ID()
+		}
+	}
+	if deviceID == "" && client.Store != nil && client.Store.ID != nil {
+		deviceID = client.Store.ID.ToNonAD().String()
+	}
+
+	chatJID := NormalizeJIDFromLID(ctx, evt.Info.Chat, client).ToNonAD().String()
+
+	websocket.Broadcast <- websocket.BroadcastMessage{
```
aldinokemal, line 169: ⚠️ This is the blocker. This is a blocking send on an **unbuffered** channel (`Broadcast = make(chan BroadcastMessage)`), and the only thing that reads from it is `RunHub()`, which we only start in REST mode (`cmd/rest.go`). 

Problem is `cobra.OnInitialize(initApp)` wires up `handleMessage` for *every* mode, and `mcp` mode auto-connects too — but nothing drains `Broadcast` there. So the first edited message that arrives in `mcp` mode will block right here forever and freeze the whatsmeow event goroutine for that device (no more messages/receipts get processed). 

Easiest fix is a non-blocking send so it just no-ops when there's no hub running:
```go
select {
case websocket.Broadcast <- msg:
default:
}
```
FWIW the existing `LOGIN_SUCCESS`/`LOGOUT_COMPLETE` sends have the same theoretical issue, but those are rare lifecycle events — this one fires on every edit, so it'll actually get hit.

src/infrastructure/whatsapp/event_message_handler.go
```diff
@@ -109,28 +112,68 @@ func handleAutoMarkRead(ctx context.Context, evt *events.Message, client *whatsm
...
-		// (HISTORY_SYNC_NOTIFICATION, APP_STATE_SYNC_KEY_SHARE, EPHEMERAL_SYNC_RESPONSE, etc.)
-		switch protocolType {
-		case "REVOKE", "MESSAGE_EDIT":
-			// These are meaningful user actions, allow webhook
-		default:
-			log.Debugf("Skipping webhook for protocol message type: %s", protocolType)
-			return
+	// Skip webhook for protocol messages that are internal sync messages.
+	// SecretEncrypted MESSAGE_EDIT has no top-level ProtocolMessage; always allow those through.
+	if !IsSecretEncryptedEdit(evt.Message) {
```
aldinokemal, line 117: minor: this guard doesn't actually change anything. A `SecretEncryptedMessage` never carries a top-level `ProtocolMessage`, so secret edits already fall straight through the filter below (that's how the webhook path works on `main`). You can drop the `IsSecretEncryptedEdit` wrapper and keep the original block as-is.

src/infrastructure/chatstorage/sqlite_repository.go
aldinokemal, line 1020: behavior change worth a second look: on `main`, a secret edit we couldn't decrypt just hit the `Skipping message - no content or media` debug log and moved on. Now any decrypt failure makes `CreateMessage` return an error → `Failed to store incoming message` gets logged for every one. 

The case where this bites: someone edits an *older* message whose message-secret isn't in our store — that'll fail to decrypt every time and spam errors for something we can't recover anyway. Might be nicer to log + skip (return nil) on `ErrSecretEditDecrypt` here instead of erroring.

src/infrastructure/chatstorage/sqlite_repository.go
```diff
@@ -1104,33 +1109,28 @@ func (r *SQLiteRepository) CreateMessage(ctx context.Context, evt *events.Messag
...
+	if key == nil || key.GetID() == "" {
+		if parsedEdit.OriginalMessageID == "" {
+			return nil
+		}
 	}
 
-	protocolMessage := evt.Message.GetProtocolMessage()
-	if protocolMessage == nil {
-		return nil
+	originalMessageID := parsedEdit.OriginalMessageID
```
aldinokemal, line 1124: small cleanup: `ExtractMessageEdit` already sets `OriginalMessageID = key.GetID()` whenever `key` is non-nil, so by the time we get here `parsedEdit.OriginalMessageID` already equals `key.GetID()`. That means this fallback branch (and the `key`-nil dance just above) can never actually run — the whole thing collapses to just `if parsedEdit.OriginalMessageID == "" { return nil }`.

aldinokemal (changes requested):
Thanks for this 🙏 The edit-detection logic itself is solid and the tests are nice. A few things before we merge though:

1. The webhook side (`message.edited` + `body` for secret edits) is actually already on `main` via #693 — that landed *after* v8.6.0, which is why you still saw the bug on 8.6.0. So that part of this PR is mostly a refactor of code that already works.
2. The genuinely new + useful bits here are **persisting** secret edits to chat storage and the **live UI update**. Worth keeping.
3. One real blocker though: the new websocket broadcast can deadlock in `mcp` mode (see inline comment on `event_message_handler.go`). That one needs a fix.

Left the details inline. Also minor: we now decrypt the same edit up to 3x per event (storage + broadcast + webhook). It's safe (decrypt is deterministic, no ratchet), just a bit wasteful — not blocking, just noting.

PR #712, opened 2026-06-07
feat: add dark mode toggle to views

rijaluddina:
This PR adds a dark/light mode toggle in the views header, with state saved in localStorage and custom dark mode styling for Fomantic UI components.


aldinokemal:
no need

PR #723, opened 2026-06-14
fix(chat): persist incoming location and link-preview messages

nazhiba:
# fix(chat): persist incoming location and link-preview messages

## Context

`ExtractMessageTextFromProto` (`src/pkg/utils/whatsapp.go`) had no case for a few message types that *do* carry displayable content, so it returned an empty string for:

- **Location pins** — `LocationMessage` (static) and `LiveLocationMessage`.
- **Bare link-preview messages** — an `ExtendedTextMessage` whose `Text` body is empty but which still has a `MatchedText` URL.

These messages also have no media type. The ingest paths skip any message with neither content nor media:

- `src/infrastructure/whatsapp/history_sync.go` (`content := utils.ExtractMessageTextFromProto(...)` then `if content == "" && mediaType == "" { continue }`)
- the equivalent skip in `src/infrastructure/chatstorage`

So today, incoming location shares (and bare link previews) are **silently dropped at ingest** — they are never stored, and therefore invisible to readers and to webhooks. This is the same class of "empty content => dropped" gap as other receive-side extraction fixes in this file.

## Fix

In `src/pkg/utils/whatsapp.go`:

- Add `FormatLocationSummary(name, address, lat, long)` which always includes a `https://maps.google.com/?q=<lat>,<long>` link, so even a coordinate-only pin yields non-empty content. Name and address are optional `—`-joined prefixes.
- Handle `GetLocationMessage()` and `GetLiveLocationMessage()` in `ExtractMessageTextFromProto` (live-location uses its caption as the name prefix).
- For `ExtendedTextMessage`, return `GetText()` when present, otherwise fall back to `GetMatchedText()` so pure link-preview messages are not empty.

No schema or migration change. Because the fix is purely at the extraction step, it also lets these message types flow through for any future re-ingest without touching storage.

## Test Results

(No screenshots — verified with unit tests only; I did not run a full live end-to-end flow.)

Added table-driven unit tests in `src/pkg/utils/whatsapp_test.go` (synthetic coordinates/names, matching the existing test style in that file):

- `TestFormatLocationSummary` — name+address, name-only, coordinates-only, and whitespace-trimmed inputs.
- `TestExtractMessageTextFromProtoLocationMessage` — `LocationMessage` with name+address, coordinate-only `LocationMessage`, `LiveLocationMessage` with caption, coordinate-only `LiveLocationMessage`.
- `TestExtractMessageTextFromProtoExtendedText` — text present wins; empty text falls back to matched URL.

Quality gates, run from `src/`:

```
gofmt -l pkg/utils/whatsapp.go pkg/utils/whatsapp_test.go   # clean (no output)
go build ./...                                              # ok
go vet ./...                                                # ok
go test ./pkg/utils/                                        # ok
```

All passing.


src/pkg/utils/whatsapp.go
```diff
@@ -99,6 +99,22 @@ func ExtractPhoneFromVCard(vcard string) string {
...
+// name and address are optional prefixes.
+func FormatLocationSummary(name, address string, lat, long float64) string {
+	var parts []string
+	if n := strings.TrimSpace(name); n != "" {
+		parts = append(parts, n)
+	}
+	if a := strings.TrimSpace(address); a != "" {
+		parts = append(parts, a)
+	}
+	mapsLink := fmt.Sprintf("https://maps.google.com/?q=%g,%g", lat, long)
```
aldinokemal, line 115: Minor: `%g` falls back to scientific notation for very small magnitudes — e.g. a coordinate of `0.00001` renders as `q=1e-05,...`, which Google Maps won't parse. This only triggers within ~1m of (0,0), so it's practically a non-issue for real GPS pins. But `%.6f` would avoid it entirely and would match the `%.6f` already used for location in `webhook_forward.go:525-537`, keeping the two representations aligned.

src/pkg/utils/whatsapp.go
```diff
@@ -212,6 +235,16 @@ func ExtractMessageTextFromProto(msg *waE2E.Message) string {
 		return "Contacts shared"
 	}
 
+	// Check for location pin
+	if loc := msg.GetLocationMessage(); loc != nil {
+		return FormatLocationSummary(loc.GetName(), loc.GetAddress(), loc.GetDegreesLatitude(), loc.GetDegreesLongitude())
```
aldinokemal, line 242: Consistency note: there's already a location-text formatter in `webhook_forward.go:517-540` that produces `Location: <name> (%.6f, %.6f)` / `Live Location: %.6f, %.6f`. With this change the same incoming location message can be rendered two different ways depending on the path (stored `Content` here vs. the webhook summary there). Not a bug, but a maintenance smell. Consider having `webhook_forward.go` call `FormatLocationSummary` too (it already takes primitives, so the layering difference — `*waE2E.Message` vs. the `interface{ GetDegreesLatitude()... }` on `data["location"]` — isn't an obstacle), or add a line to the PR explaining why both formats coexist.

aldinokemal (commented):
## Code review

**Solid, surgical fix.** I verified the problem statement against the codebase: `ExtractMessageTextFromProto` had no case for location/live-location or bare link-preview messages, and both ingest paths drop messages where `content == "" && mediaType == ""` (`history_sync.go:238`, `chatstorage/sqlite_repository.go:1344`), so these were silently dropped at ingest. The diagnosis is accurate and the fix is at the right layer.

**Things I confirmed are correct:**
- No regression in the `ExtendedTextMessage` change — when both `Text` and `MatchedText` are empty it now falls through to the remaining checks and still returns `""` (an extended-text message won't match image/video/etc.), so behavior is equivalent in that case and strictly better when a `MatchedText` exists.
- Location checks are placed just before the final `return ""`, not interfering with the caption checks above.
- Style matches `FormatContactSummary`; test helpers match existing style; `f64Ptr` is new (no duplicate), `strPtr` is reused.
- Good unit coverage for the extraction layer.

**Follow-ups (see inline comments):**
1. **Duplicate, divergent location formatting** — there's already a location formatter in `webhook_forward.go:517-540` producing `Location: <name> (%.6f, %.6f)`. The same incoming message can now render two different ways depending on path. Worth consolidating, or at least noting why two formats coexist.
2. **`%g` can emit scientific notation** for coordinates very near 0 (e.g. `0.00001` → `1e-05`), which Google Maps won't parse. Practically irrelevant for real GPS pins, but `%.6f` (as `webhook_forward.go` already uses) avoids it and aligns the two formats.

Neither is blocking. Posting as a comment rather than an approval since #1 is a real consistency follow-up worth a maintainer decision.

aldinokemal:
### Follow-up: the "invisible to webhooks" premise needs correcting, and the scope is narrower than stated

I traced the full receive path and one of the stated justifications doesn't hold up.

`handleMessage` (`event_message_handler.go:45-60`) does two **independent** things:

1. **Stores** the message → `chatStorageRepo.CreateMessage` → hits the `content == "" && mediaType == "" → skip` path. Location pins are dropped here. ✅ accurate.
2. **Forwards to webhook** → `handleWebhookForward` → `forwardMessageToWebhook` → `buildOtherMessageTypes` sets `payload["location"]` / `payload["live_location"]` directly from the proto (`event_message.go:404-409`), and `webhook_forward.go:517-540` already builds a `Location: …` / `Live Location: …` text summary.

The webhook forward does **not** depend on storage succeeding, so:

- **Webhooks already receive locations.** The PR description's "never stored, and therefore invisible to readers **and to webhooks**" is incorrect for the webhook half.
- **Chatwoot already receives them too**, since it's wired through the same `handleWebhookForward` path (`config.ChatwootEnabled`, `event_message_handler.go:166`).

So the **only** real gap this PR closes is the **locally-stored chat history** — i.e. the REST endpoints that read messages back out of the chat store. That's a legitimate completeness fix, but narrower than the description implies.

Two asks before merge:

1. **Update the PR/commit description** — drop the "invisible to webhooks" rationale; it'll otherwise be inaccurate in the history. The accurate framing is "persisted to stored chat history so it surfaces via the read API."
2. **Reconsider the `MatchedText` fallback for `ExtendedTextMessage`.** A user-sent link produces an `ExtendedTextMessage` whose `Text` *is* the typed URL (non-empty), so `GetText()` already returns it. An empty `Text` with only `MatchedText` set is hard to produce in practice — that half looks like effectively dead code. Either drop it or include a concrete repro of a message that hits it.

The location fix itself is good; this is about right-sizing the claim and the surface area (plus the `FormatLocationSummary` / `webhook_forward.go` duplicate-formatter note from the inline review).

aldinokemal:
Thanks @nazhiba — this fix landed on main via #722 (`5aea840`) with the tests in `b58c932`, so the diff here is now empty. Closing as already shipped; credit for the diagnosis and fix approach is much appreciated.

PR #728, opened 2026-06-16
fix(device): logout should disconnect session, not delete device entry

huboperacional:
## Problem

When a user clicks **Logout** in the dashboard (or calls `POST /devices/:id/logout`), the device record is permanently deleted. This means the device name and slot are lost — the user has to re-add the device and rename it every time they log out of WhatsApp.

This is inconsistent with how WhatsApp Web itself behaves: logging out of a linked device disconnects the session but keeps the device entry in the list (ready to re-link by scanning a new QR code). The current behaviour conflates *logout* (disconnect session) with *remove* (delete slot).

## Root cause

`AppLogout` → `serviceApp.Logout()` → `PurgeDevice()` → `RemoveDevice()` (deletes the registry record). In addition, the `SetOnLoggedOut` callback also calls `RemoveDevice()` when WhatsApp sends a remote unlink event.

## Fix

Introduce two distinct operations:

| Operation | What it does |
|---|---|
| `LogoutDeviceKeepSlot` | Calls `cli.Logout()` + `cli.Disconnect()`, resets the in-memory client state, clears the JID — but **does NOT delete the registry record** |
| `PurgeDevice` (existing, via DELETE) | Full removal: calls `LogoutDeviceKeepSlot` first, then deletes the registry record |

`SetOnLoggedOut` callback now calls `resetDeviceKeepSlot` (keep slot) instead of `RemoveDevice` (delete slot).

A new WebSocket event `DEVICE_LOGGED_OUT` is broadcast on logout so the frontend refreshes the device list without clearing the current selection (unlike `DEVICE_REMOVED`).

## Changes

- `src/infrastructure/whatsapp/device_instance.go` — add `ResetClient()` method (clears client, JID, phone, state)
- `src/infrastructure/whatsapp/device_manager.go` — add `LogoutDeviceKeepSlot` / `resetDeviceKeepSlot`; change `SetOnLoggedOut` callback
- `src/usecase/app.go` + `src/usecase/device.go` — `Logout()` delegates to `LogoutDeviceKeepSlot`; `RemoveDevice()` delegates to `PurgeDevice()`
- `src/views/index.html` — handle `DEVICE_LOGGED_OUT` WebSocket event (refresh list, keep selection)

## Testing

**CI (GitHub Actions):** go build ✅ · go vet ✅ · go test ✅ · Docker image built and pushed to GHCR.

**Runtime (deployed image on Docker Swarm):**
1. `POST /devices` → created slot `08617698-d64b-4436-a41b-190945fed8d9`
2. Confirmed device appears in `GET /devices`
3. `POST /devices/08617698-.../logout` → 200 "Logout requested"
4. `GET /devices` → device **still present** ✅ (fix verified)
5. `DELETE /devices/08617698-...` → 200 "Device removed"
6. `GET /devices` → device **gone** ✅ (delete semantics correct)

Chat history is intentionally preserved on logout (whatsmeow stores history locally in SQLite); it is cleared on `DELETE` / full purge.


src/infrastructure/whatsapp/device_manager.go
```diff
@@ -222,6 +222,61 @@ func (m *DeviceManager) PurgeDevice(ctx context.Context, deviceID string) error
...
+		return fmt.Errorf("device id is required")
+	}
+
+	inst, ok := m.GetDevice(deviceID)
+	if !ok || inst == nil {
+		return fmt.Errorf("device %s not found", deviceID)
+	}
+
+	var firstErr error
+	if cli := inst.GetClient(); cli != nil {
```
aldinokemal, line 253: This only clears the session when there is a live logged-in client. If the slot was loaded from storage and has no client yet, the old whatsmeow row stays around and can get matched back on restart. We should delete the stored JID/session/key rows before clearing the slot JID.

src/infrastructure/whatsapp/device_manager.go
```diff
@@ -523,7 +578,9 @@ func (m *DeviceManager) EnsureClient(ctx context.Context, deviceID string) (*Dev
 	})
 
 	inst.SetOnLoggedOut(func(deviceID string) {
-		m.RemoveDevice(deviceID)
+		// On remote logout (device unlinked from the phone) keep the slot so it can
```
aldinokemal, line 632: This fixes the lazy EnsureClient logout path, but the startup InitWaCLI path still registers dm.RemoveDevice for remote logout. So a logout from the phone can still remove the slot. Let's route every LoggedOut callback through the same keep-slot reset behavior.

src/usecase/device.go
```diff
@@ -56,11 +57,17 @@ func (s *serviceDevice) AddDevice(ctx context.Context, deviceID string) (*domain
...
+func (s *serviceDevice) RemoveDevice(ctx context.Context, deviceID string) error {
 	if s.manager == nil {
 		return fmt.Errorf("device manager not initialized")
 	}
-	s.manager.RemoveDevice(deviceID)
+	// Deleting a device fully purges it: logs it out of WhatsApp, clears its session
+	// and chat data, then removes the slot from the registry. (Logout, in contrast,
+	// keeps the slot.) Purge removes the slot even if the WhatsApp logout reports
+	// warnings (e.g. an already-dead session), so don't fail the delete on those.
+	if err := s.manager.PurgeDevice(ctx, deviceID); err != nil {
```
aldinokemal, line 74: This always returns success even if purge failed to delete chat data, store rows, or key rows. Since DELETE now promises a real purge, we should surface local cleanup failures and only treat the WhatsApp unlink call as best-effort if that is intentional.

src/views/index.html
```diff
@@ -409,6 +409,20 @@ <h1 class="ui header center aligned">
                             }
                             break;
                         }
+                        case 'DEVICE_LOGGED_OUT': {
```
aldinokemal, line 415: This new logout event is fine for logout, but the delete modal still fire-and-forgets /app/logout before DELETE. That can race this event against the delete response and make a removed slot show up again. Better to let DELETE own the purge/logout flow.

src/infrastructure/whatsapp/device_manager.go
```diff
@@ -222,6 +222,61 @@ func (m *DeviceManager) PurgeDevice(ctx context.Context, deviceID string) error
 	return firstErr
 }
 
+// LogoutDeviceKeepSlot logs the device out of WhatsApp (clearing its session/keys)
+// but PRESERVES the device slot in the registry, so it keeps its id and display name
+// and can be re-paired later under the same id. Unlike PurgeDevice, it does not remove
+// the device record. Removing the slot entirely is the job of RemoveDevice (DELETE).
+func (m *DeviceManager) LogoutDeviceKeepSlot(ctx context.Context, deviceID string) error {
```
aldinokemal, line 243: This is core device lifecycle behavior, so it really needs regression coverage. At minimum I'd want tests for keep-slot logout, no-client logout after reload, remote logout, DELETE purge by slot/JID, and separate keys DB cleanup.

aldinokemal (dismissed):
Not merge-ready yet. The keep-slot direction is good, but a few lifecycle paths can still leave old sessions around or make the device list drift.

huboperacional:
**Round 2 — pushed in `43f251d`. Thanks for the thorough review; all points addressed.**

**@aldinokemal**
- **`device_manager.go:240` (orphan rows on no-client logout):** `LogoutDeviceKeepSlot` now resolves the JID *before* the reset clears it and always calls a new `deleteStoreRowsForJID` helper, which deletes the whatsmeow rows from **both** the primary and keys containers by **NonAD** JID (mirroring `LoadExistingDevices`). This also fixes a latent bug: the old `PurgeDevice` compared the **AD** JID string to the slot id, so it never matched and never deleted the rows.
- **`device_manager.go:581` (two callback paths):** both the startup `InitWaCLI` path and the lazy `EnsureClient` path now route through the same `keepSlotLogout`, so a remote logout never deletes the slot.
- **`device.go:68` (DELETE masking failures):** `RemoveDevice` now returns the purge error. Inside `PurgeDevice`, the WhatsApp unlink is explicitly best-effort (a dead session is logged, not fatal), while **local** cleanup failures (chatstorage / store / keys) are surfaced — matching exactly your "best-effort unlink, surface local cleanup" request.
- **`index.html:412` (delete race):** the delete modal no longer fires `/app/logout` before `DELETE`; `DELETE` owns the full purge/logout flow.
- **`device_manager.go:229` (tests):** added regression coverage for keep-slot logout, no-client logout after reload, remote logout, DELETE purge by slot/JID, separate keys DB cleanup, plus persistence-error propagation and the empty-JID no-op.

**@coderabbitai**
- `resetDeviceKeepSlot` now returns an `error` and propagates `SaveDeviceRecord` failures; `LogoutDeviceKeepSlot` joins them via `errors.Join` (your suggested form).
- `RemoveDevice` returns `fmt.Errorf("purge device %s: %w", …)` instead of `nil`.
- openapi: `DELETE /devices/{id}` → `{device_id}`.
- **validate-first:** done in `Logout` / `LogoutDevice` / `RemoveDevice` via a new `ValidateDeviceID`. On **`ClientFromContext`** I deliberately did *not* apply it here: the `/devices/{device_id}` routes are registered on `apiGroup` **without** `DeviceMiddleware` (only `headerDeviceGroup` has it, in `cmd/rest.go`), so there's no device in context — `ClientFromContext` would fall back to the global client (wrong for multi-device) and these operations must also work with no live client (the no-client logout case above). So I validate the id and operate via the manager by id.

**@kilo-code-bot**
- `handleLoggedOut` no longer truncates chat history on remote logout (history is preserved, matching keep-slot semantics) and now broadcasts `DEVICE_LOGGED_OUT` ("slot kept") instead of the misleading `LOGOUT_COMPLETE` "device removed from server".

Compiled + `go vet` + `go test` + image build all green via a temp CI workflow on a throwaway branch (the upstream repo doesn't run Go checks on PRs).

huboperacional:
@aldinokemal addressed all points in 43f251d (summary above); CodeRabbit re-approved. Whenever you have a moment 🙏

aldinokemal (approved):
Round-3 verification: all six round-1 findings confirmed fixed with SQLite-backed regression tests. Two residual issues found in a final review pass were fixed in e4f3049: (1) keepSlotLogout now falls back to NonAD-JID resolution when the LoggedOut callback holds a stale AD-JID id after loadFromRegistry re-keys the slot, so remote logout can no longer leave a stale-JID slot and an orphan keys row; (2) the WhatsApp unlink is attempted whenever the client is paired (Store.ID set), not only while connected, so a momentarily-offline logout no longer leaves the device linked on the phone forever. go build / go vet / go test ./... all pass.

PR #756, opened 2026-07-04
feat(webhook): ignore group messages per instance, not just globally

huboperacional:
## Problem

`WHATSAPP_WEBHOOK_IGNORE_JIDS` (added in #736) is a single **global**, boot-time-only env var: it applies to every device the same way, and changing it requires editing `.env` + restarting the container. There's no way to mute group traffic on just one instance while keeping it on another, and no runtime API/UI to flip it.

Since #736 merged, per-device webhook configuration landed (`d1ac267`): `DeviceWebhookConfig` persists `webhook_url`/`webhook_secret`/`webhook_events`/`webhook_insecure_skip_verify` per device, with `PATCH`/`GET /devices/:device_id/webhook` reading/writing it at runtime, no restart needed. This PR extends that same infrastructure rather than building something parallel.

## Solution

Add `webhook_ignore_groups` (`*bool`, nullable) to the per-device webhook config:

- **`nil`** (never configured) → falls back to today's behavior: checks whether the global `WHATSAPP_WEBHOOK_IGNORE_JIDS` list contains the `@g.us` wildcard. Existing users of the env var see zero behavior change.
- **`true`/`false`** (explicitly set) → becomes the **sole source of truth** for group JIDs on that device, overriding the global `@g.us` wildcard either way (even if the device opts to keep receiving groups despite a global mute, or vice versa).
- Non-group JIDs are untouched — the global list remains the only mechanism there, exactly as before.

No new persistence mechanism, no new route: reuses `PATCH`/`GET /devices/:device_id/webhook` as-is, just with one more optional field.

## Changes

- `domains/chatstorage/chatstorage.go` — `WebhookIgnoreGroups *bool` on `DeviceRecord` and `DeviceWebhookConfig`.
- `infrastructure/chatstorage/sqlite_repository.go` — Migration 35 (`ALTER TABLE devices ADD COLUMN webhook_ignore_groups BOOLEAN DEFAULT NULL`); `GetDeviceRecordByJID`/`SetDeviceWebhookConfig`/`GetDeviceWebhookConfig` read/write the field (no `COALESCE`, so explicit `false` never collapses to `nil`).
- `infrastructure/whatsapp/webhook_forward.go` — `shouldIgnoreWebhookJID` gains `deviceIgnoreGroupsOverride` (per-device lookup) and `globalIgnoresAllGroups` (fallback check) helpers; device override only affects `@g.us`-suffixed JIDs, precedence exactly as described above. `getWebhookConfigForDevice` also propagates the field for consistency with the other webhook settings it already returns.
- `ui/rest/device.go` — `PATCH`/`GET /devices/:device_id/webhook` accept/return `webhook_ignore_groups` (serializes as JSON `null` when never configured).
- Tests across all four packages: persistence round-trip (including the explicit-`false`-must-not-collapse-to-`nil` case), the full precedence matrix in the resolver (device `true`/`false`/`nil` × global list empty/`@g.us`/mixed, plus a non-group-JID isolation test), and the REST handlers (forwarding + response shape for `true`/`false`/never-configured).

## Testing

No local Go toolchain on the machine that authored this, so `gofmt`/`go build`/`go vet`/`go test` were run via a disposable CI workflow pushed to a throwaway branch on the fork (not part of this PR) — `gofmt` (changed files), `go build ./...`, `go vet ./...`, and `go test` on `infrastructure/chatstorage`, `infrastructure/whatsapp`, `ui/rest`, plus the full suite — all green, confirmed twice (once initially, once after fixing a small consistency gap a final review caught in `getWebhookConfigForDevice`).

Not yet smoke-tested against a live WhatsApp device/group — happy to do that before merge if useful.

### DIFF
--- a/src/infrastructure/whatsapp/webhook_forward.go
+++ b/src/infrastructure/whatsapp/webhook_forward.go
@@ -207,29 +207,74 @@ func isEventWhitelistedForDevice(eventName string, deviceConfig *domainChatStora
 
 // shouldIgnoreWebhookJID reports whether an event should be skipped for WHATSAPP_WEBHOOK
 // forwarding because its chat or sender JID matches WHATSAPP_WEBHOOK_IGNORE_JIDS (e.g. the
-// "@g.us" wildcard to drop all group traffic). The JID fields live in the nested inner
-// payload, so it descends one level. Both the resolved phone JIDs (chat_id/from) and the
-// LID forms (chat_lid/from_lid) are matched: a LID-migrated event keeps the @lid JID in the
-// *_lid fields while chat_id/from hold the resolved phone JID, so an "@lid" pattern (or an
-// exact ...@lid) only matches via the *_lid fields. It is a no-op when the ignore list is
-// empty, the inner payload is absent, or no JID matches — so events without a JID and the
-// default (no list configured) keep forwarding unchanged. This only gates the generic
-// webhook; the Chatwoot path keeps its own CHATWOOT_IGNORE_JIDS filter.
+// "@g.us" wildcard to drop all group traffic), OR because the originating device has an
+// explicit per-device override for group messages (webhook_ignore_groups, set via
+// PATCH /devices/:device_id/webhook). The device override takes precedence over the
+// global "@g.us" wildcard specifically for group JIDs; it has no effect on non-group JIDs,
+// where the global list remains the only mechanism (unchanged from before this feature).
+// The JID fields live in the nested inner payload, so it descends one level. Both the
+// resolved phone JIDs (chat_id/from) and the LID forms (chat_lid/from_lid) are matched.
+// It is a no-op when the inner payload is absent or no JID matches.
 func shouldIgnoreWebhookJID(payload map[string]any) bool {
-	ignore := config.WhatsappWebhookIgnoreJids
-	if len(ignore) == 0 {
-		return false
-	}
 	data, ok := payload["payload"].(map[string]any)
 	if !ok {
 		return false
 	}
+
+	deviceJID, _ := payload["device_id"].(string)
+	groupOverride := deviceIgnoreGroupsOverride(deviceJID)
+	ignore := config.WhatsappWebhookIgnoreJids
+
 	for _, key := range []string{"chat_id", "from", "chat_lid", "from_lid"} {
-		if jid, _ := data[key].(string); utils.MatchesIgnoredJID(jid, ignore) {
+		jid, _ := data[key].(string)
+		if jid == "" {
+			continue
+		}
+		if strings.HasSuffix(jid, "@g.us") {
+			if groupOverride != nil {
+				if *groupOverride {
+					return true
+				}
+				continue
+			}
+			if globalIgnoresAllGroups(ignore) {
+				return true
+			}
+			continue
+		}
+		if utils.MatchesIgnoredJID(jid, ignore) {
 			return true
 		}
 	}
 	return false
+}
+
+// globalIgnoresAllGroups reports whether the global WHATSAPP_WEBHOOK_IGNORE_JIDS list
+// contains the "@g.us" wildcard (ignore all groups). Used as the fallback when a device
+// has no explicit webhook_ignore_groups override.
+func globalIgnoresAllGroups(ignore []string) bool {
+	for _, pattern := range ignore {
+		if strings.TrimSpace(pattern) == "@g.us" {
+			return true
+		}
+	}
+	return false
+}
+
+// deviceIgnoreGroupsOverride returns the per-device override for ignoring group messages
+// in webhook forwarding (webhook_ignore_groups), or nil when the device has never set it
+// -- the caller falls back to the global "@g.us" wildcard. Uses the same test seam
+// (getDeviceRecordForTest) as getWebhookConfigForDevice, so existing tests that don't stub
+// it keep seeing nil (unchanged behavior).
+func deviceIgnoreGroupsOverride(deviceJID string) *bool {
+	if deviceJID == "" {
+		return nil
+	}
+	record, err := getDeviceRecordForTest(deviceJID)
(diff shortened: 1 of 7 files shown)


huboperacional:
Thanks for the reviews (@coderabbitai, and the Codex/Kilo bots) — both findings were real. Addressed in two follow-up commits:

## 1. Exact/specific group JID entries no longer filtered (regression) — fixed in `c90ec60`

Flagged on `webhook_forward.go` (and the paired missing-test note). Correct catch: the nil-override fallback only checked the literal `@g.us` wildcard via `globalIgnoresAllGroups`, so an operator who put a *specific* group JID (e.g. `120363999000111@g.us`) in `WHATSAPP_WEBHOOK_IGNORE_JIDS` would silently stop having it filtered — a regression vs. the pre-existing behavior.

Fix: in the `@g.us` branch with no per-device override, fall through to `utils.MatchesIgnoredJID(jid, ignore)` (matches both the `@g.us` wildcard **and** exact JIDs), restoring the original behavior. `globalIgnoresAllGroups` was now unused and removed. Added `TestWebhookIgnoreJID_NilDeviceOverrideRespectsExactGroupJIDMatch` covering the exact-group-JID-in-global-list case.

## 2. PATCH silently resets `webhook_ignore_groups` to nil when omitted — fixed in `197d39b`

Also a real footgun. `UpdateDeviceWebhook` now reads the existing config and preserves the stored `webhook_ignore_groups` when the field is omitted from the request body (`nil`), instead of overwriting it. Explicit `true`/`false` still overwrite — the `*bool` lets us distinguish "sent false" from "omitted". This is your proposed handler-level fix; keeping the merge at the handler leaves `SetDeviceWebhookConfig` a faithful writer.

Trade-off we accepted deliberately: there's no JSON value that resets the field back to `nil` ("revert to global default") via this endpoint. In practice that's a non-issue — when the global `WHATSAPP_WEBHOOK_IGNORE_JIDS` isn't in play for a device, `false` and `nil` behave identically (both forward groups), and the per-device override is meant to be the authoritative, sticky control once set.

Added `TestUpdateDeviceWebhook_PreservesIgnoreGroupsWhenOmitted` and `TestUpdateDeviceWebhook_ExplicitFalseOverridesExisting`.

Verified via `gofmt` / `go build ./...` / `go vet ./...` / `go test ./...` (all packages + full suite), all green.

huboperacional:
Rebased onto v9.0.0. Two adjustments were needed beyond the mechanical rebase:

- `c.UserContext()` -> `c.Context()` for the Fiber v3 migration (#764)
- the migration moved from 35 to 44 - 35 is now `ad_jid` from #762, and the runner is index-based, so appending at the end is the only safe position

`GetDeviceRecordByJID` was rewritten upstream (LIMIT 2 + ambiguity guard); I kept your version and just threaded the new column through the SELECT and Scan. Build, vet and the full test suite pass.

src/ui/rest/device.go
```diff
@@ -229,14 +230,23 @@ func (handler *Device) UpdateDeviceWebhook(c fiber.Ctx) error {
 		})
 	}
 
+	existing, err := handler.Service.GetDeviceWebhookConfig(c.Context(), deviceID)
+	utils.PanicIfNeeded(err)
+
+	ignoreGroups := req.WebhookIgnoreGroups
+	if ignoreGroups == nil && existing != nil {
```
aldinokemal, line 237: [Major] Explicit `null` cannot restore the global fallback. JSON binding produces a nil `*bool` for both an omitted field and `"webhook_ignore_groups": null`, and this branch always carries the persisted value forward. Once a device sets `true` or `false`, there is therefore no API operation that writes SQL `NULL` again, even though `NULL` is the documented inheritance state and GET exposes it. Track field presence separately (for example with a custom nullable type, `json.RawMessage`, a map, or `**bool`) so omitted means “preserve” while explicit `null` means “clear”, and add a regression test covering `true -> null -> nil`.

src/infrastructure/whatsapp/webhook_forward.go
```diff
@@ -207,31 +208,64 @@ func isEventWhitelistedForDevice(eventName string, deviceConfig *domainChatStora
...
+		jid, _ := data[key].(string)
+		if jid == "" {
+			continue
+		}
+		if strings.HasSuffix(jid, "@g.us") {
+			if groupOverride != nil {
+				if *groupOverride {
+					return true
+				}
+				continue
```
aldinokemal, line 239: [Major] `false` currently disables exact global group exclusions too. This `continue` skips `MatchesIgnoredJID`, so if the same group is explicitly listed in `WHATSAPP_WEBHOOK_IGNORE_JIDS` (for example `120363...@g.us`), the event is forwarded. The documented contract says the per-device flag overrides the global `@g.us` all-groups wildcard, not exact per-group entries. Separate wildcard detection from exact matching: an explicit `false` should suppress only the wildcard while still honoring an exact group JID, and add a regression case for `false` plus the exact same group in the global list.

aldinokemal (commented):
I reviewed the per-device webhook filtering flow end to end: nullable persistence/migration behavior, PATCH/GET semantics, global-vs-device precedence, resolver behavior, and the added tests.

I found two correctness gaps that should be addressed before merge:

1. The API cannot restore `webhook_ignore_groups` to SQL `NULL` after a device sets `true` or `false`, because omitted and explicit JSON `null` collapse to the same Go value and are always treated as “preserve existing”.
2. An explicit device override of `false` also bypasses exact group-JID entries in `WHATSAPP_WEBHOOK_IGNORE_JIDS`, although the documented contract says it should override only the global `@g.us` wildcard.

I left inline comments with concrete fixes and requested regression cases. The migration numbering is currently safe against `main` (this becomes migration 44), but this branch is presently not mergeable and there are no pull-request workflow runs attached to the current head. After fixing the cases below, please rebase and run `gofmt`, `go build ./...`, `go vet ./...`, and `go test ./...`.

PR #757, opened 2026-07-07
feat(chatwoot): quoted reply threading in both directions + group name resolution

lovaraju987:
## Summary

Fixes **5 related bugs** that broke Chatwoot ↔ WhatsApp reply threading. The first four break quoted replies in both directions; the fifth fixes group participant name resolution.

Without these fixes, reply bubbles either don't appear at all or appear as standalone messages in Chatwoot.

### 1. `event_message_handler.go:177` — Live webhooks never created `chatwoot_message_links`

The webhook goroutine used `context.Background()` instead of inheriting from the parent `ctx`, so `DeviceFromContext` returned `nil` and links were silently skipped.

**Before:**
```go
webhookCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```

**After:**
```go
webhookCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
```

Only affected live inbound webhooks; history sync was unaffected because it uses a different code path.

### 2. `send.go mergeReplyContext()` — Cross-client quote rendering

WhatsApp iOS/Android 2.24+ render the quote bubble from `ExtendedTextMessage`; older Web uses `Conversation`. Setting both covers the full client matrix.

### 3. `webhook_forward.go syncPayloadToChatwoot()` — Inbound reply ID translation

Chatwoot's `in_reply_to` expects **its own message ID**, not the WhatsApp message ID. GOWA was sending the WA ID (as `in_reply_to_external_id`) but not translating it to the Chatwoot ID, so the reply showed as standalone.

The fix looks up the WA ID in `chatwoot_message_links` and sets `in_reply_to` to the Chatwoot ID, falling back gracefully when no link exists.

### 4. `chatwoot.go` — New outbound reply handler

`resolveChatwootReplyMessageID()` reads `content_attributes.in_reply_to` from the Chatwoot webhook, looks up the WA message ID via the link table, and sets `ReplyMessageID` on the outgoing SendText request. Agent replies in Chatwoot now thread as quoted bubbles on WhatsApp.

### 5. `event_message.go` — Group participant name lookup

For group messages where `PushName` is empty in the event, `resolveGroupParticipantDisplayName()` looks up the sender's display name from the group participant metadata. Chatwoot now shows names instead of phone numbers in group replies.

## Testing

All five fixes have been tested end-to-end in production at OneTo7 Solutions:
- **Outbound** (Chatwoot → WhatsApp): agent reply arrives as quoted bubble ✅
- **Inbound** (WhatsApp → Chatwoot): customer reply shows as linked reply ✅
- **Groups**: reply threading works and participant names resolve correctly ✅

## Notes

- Fix 1 is a pre-existing GOWA bug that affected every live inbound webhook
- Fixes 2 and 3 are pure enhancements; no behavior change for existing users
- Fixes 4 and 5 are new features
- Backwards-compatible: all lookups fall back to existing behavior when no link/name is found


aldinokemal (commented):
Reviewed the current head as a Go/whatsmeow change. I found blocking correctness issues, so this is not ready to merge yet.

1. **P1 — current head does not compile:** `types.GroupParticipant.PhoneNumber` is a `types.JID` value, but `resolveGroupParticipantDisplayName` compares it with `nil`. The existing unresolved inline thread at `event_message.go:476` is still valid; use `!p.PhoneNumber.IsEmpty()`.
2. **P1 — reply lookup is not scoped to the destination chat:** `resolveChatwootReplyMessageID` receives `chatJID` but never validates it. The repository lookup is scoped only by device and Chatwoot message ID, while `mergeReplyContext` later loads the WhatsApp message by device + message ID and embeds its stored sender/content into the quote. A stale, colliding, or malformed `in_reply_to` can therefore bind a message from another conversation on the same device and quote the wrong content/participant. I left an inline comment with the expected guard.
3. **P2 — attachment replies still bypass reply context:** the attachment branch returns before `ReplyMessageID` is resolved and passed to image/video/audio/file requests. The existing unresolved thread at `chatwoot.go:593` remains valid.
4. **P2 — fast back-to-back inbound replies can lose threading:** forwarding is concurrent, so the quoted message's link may not exist when `buildInReplyToAttrs` runs. The existing unresolved thread at `webhook_forward.go:1037` should be addressed by ordering or retrying when the link is temporarily missing.

There is no GitHub Actions workflow run for head `334320fb474e3ceb7a44d08b2b4076a997ed371a` (only the CodeRabbit status), and GitHub currently reports the PR as not mergeable. Please rebase, fix the compile error and scoping issues, then run `gofmt`, `go vet ./...`, and `go test ./...`, including tests for cross-conversation reply IDs and attachment replies.

src/ui/rest/chatwoot.go
```diff
@@ -135,6 +135,49 @@ func chatwootLinkChatJID(destination string) string {
...
+			inReplyTo = int64(n)
+		case float64:
+			inReplyTo = int64(n)
+		}
+	}
+	if v, ok := ca["in_reply_to_external_id"].(string); ok {
+		inReplyToExternalID = v
+	}
+	if inReplyTo == 0 && inReplyToExternalID == "" {
+		return nil
```
aldinokemal, line 159: **P1 — scope the resolved reply to the current WhatsApp chat.** `chatJID` is passed into this helper but is never used. This lookup only scopes by `deviceID` + Chatwoot message ID; the returned link already contains `WhatsAppChatJID`. Later, `mergeReplyContext` looks up the WhatsApp message by device + message ID and embeds its stored sender/content into the quote. A stale/colliding `in_reply_to` from another conversation on the same device can therefore quote the wrong chat's message. Before returning the ID, normalize and require `link.WhatsAppChatJID == chatJID`; otherwise omit the reply context. The external-ID fallback should receive the same chat-scoping treatment rather than accepting an arbitrary WAID.

PR #765, opened 2026-07-13
feat: cross-origin enablers for the standalone gowa-ui dashboard

aldinokemal:
## Summary

Non-breaking groundwork so a standalone web UI ([gowa-ui](https://github.com/aldinokemal/gowa-ui)) hosted on any origin can talk to this server. The embedded dashboard keeps working unchanged.

- **CORS**: moved before the static routes (so `/statics` media and QR PNGs carry CORS headers) and widened `AllowHeaders` with `Authorization` + `X-Device-Id`; explicit `AllowMethods`. Wildcard origin stays safe because auth is header-based, not cookie-based.
- **`GET /app/info`**: authenticated JSON endpoint exposing version, OS name, base path, media size limits, and `chatwoot_enabled` — replaces the values the embedded UI received via Go template injection. Documented in `docs/openapi.yaml`.
- **WebSocket query auth**: browsers cannot set an `Authorization` header on WebSocket connections, so `/ws` now also accepts `?authorization=<base64(user:pass)>` — a small middleware restores it into the header before the standard basic-auth check runs (use TLS in production; documented tradeoff).

## Verification

- `go vet ./...` and `go test ./...` green; new table tests cover the CORS preflight header set, `/app/info` auth behaviour, and the WS query-auth middleware (incl. the fasthttp `+`→space decoding edge).
- Verified live with curl: preflight returns the widened header set; `/app/info` gates 401/200; wrong WS query creds → 401, correct creds reach the device layer; statics serve pre-auth with CORS headers.
- Embedded UI regression-checked (login + send still work same-origin).

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -8,6 +8,10 @@ info:
     Device scoping:
     - Send `X-Device-Id` on all device-scoped REST calls.
     - WebSocket: connect to `/ws?device_id=<id>`.
+    - WebSocket auth (when basic auth is enabled): browsers cannot set an
+      Authorization header on WebSocket connections, so also pass
+      `&authorization=<base64(user:pass)>` in the query string. Use TLS in
+      production since the credential is visible in the URL.
 servers:
   - url: http://localhost:3000
 tags:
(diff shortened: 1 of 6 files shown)


aldinokemal:
Consolidated into #766 (single unified branch/PR per maintainer preference) — all commits of this branch are part of feat/runtime-ui-cutover.

PR #767, opened 2026-07-13
fix: sent messages missing from chat storage under write contention

aldinokemal:
## Symptom

Send a message (e.g. from the dashboard or the REST API) while a freshly-paired device is still processing history sync: the chat's `last_message_time` updates, but `GET /chat/{jid}/messages` returns 0 messages — the sent message reached WhatsApp but never landed in chat storage.

## Root cause

`wrapSendMessage` stores the sent message in a detached goroutine with a **2-second** budget, while the chat-storage SQLite connection uses `busy_timeout=30000` — so during history-sync write bursts a queued write happily waits longer than the whole budget. Worse, `StoreSentMessageWithContext` wrote the **chat row first** and checked the context deadline *between* the two writes, so the common partial failure was exactly the misleading state above: bumped chat, missing message, and only a server-side warn log.

## Fix

- Store the message row **before** the chat bump — a partial failure now loses only the `last_message_time` update, which is invisible and self-heals on the next message.
- Raise the detached store budget from 2s to 15s (still bounded, comfortably under `busy_timeout`).
- Failure logs now include the message ID and recipient so drops are diagnosable.

## Verification

`go vet` and the full `go test ./...` suite pass. Reproduced the original symptom on a live server (paired device mid-history-sync, sent to a group → chat bumped to *now*, 0 messages; `Timeout storing sent message` in logs matches the 2s expiry).

Independent of #765/#766 — can merge in any order.

### DIFF
--- a/src/infrastructure/chatstorage/sqlite_repository.go
+++ b/src/infrastructure/chatstorage/sqlite_repository.go
@@ -2203,15 +2203,42 @@ func (r *SQLiteRepository) StoreSentMessageWithContext(ctx context.Context, mess
 	normalizedJID := whatsapp.NormalizeJIDFromLID(ctx, jid, client)
 	chatJID := normalizedJID.String()
 
+	// Extract media info from the protobuf message if available
+	var mediaType, filename, mediaURL, directPath string
+	var mediaKey, fileSHA256, fileEncSHA256 []byte
+	var fileLength uint64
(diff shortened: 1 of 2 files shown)


aldinokemal:
Consolidated into #766 — the fix is merged into feat/runtime-ui-cutover.

PR #771, opened 2026-07-14
fix: retry whatsmeow queries on websocket disconnect

shaburi:
Issues:
The server occasionally throws 500 or 502 errors ("websocket disconnected 
before info query returned response") when fetching group info.

Cause:
WhatsApp's servers routinely close idle websockets. If a GetGroupInfo query 
fires right as the socket drops, it fails immediately. The whatsmeow library 
auto-reconnects within ~1s, but the app code does not retry the failed query.

Proposed Fix:
Added a GetGroupInfoWithRetry helper in utils/whatsapp.go. If GetGroupInfo 
fails and the client is disconnected, it waits 2s for auto-reconnect, then 
retries once. All 5 call sites updated.

### DIFF
--- a/src/infrastructure/chatwoot/sync.go
+++ b/src/infrastructure/chatwoot/sync.go
@@ -93,7 +93,7 @@ func (g *groupNameResolver) resolve(ctx context.Context, waClient *whatsmeow.Cli
 	}
 	lookupCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
 	defer cancel()
-	info, err := waClient.GetGroupInfo(lookupCtx, jid)
+	info, err := utils.GetGroupInfoWithRetry(lookupCtx, waClient, jid)
 	if err != nil || info == nil || info.Name == "" {
 		return ""
 	}
--- a/src/infrastructure/whatsapp/webhook_forward.go
+++ b/src/infrastructure/whatsapp/webhook_forward.go
@@ -1282,7 +1282,7 @@ func getGroupName(ctx context.Context, groupJID string) string {
 	defer cancel()
 
 	logrus.Debugf("Chatwoot: Fetching group info for %s", groupJID)
-	groupInfo, err := client.GetGroupInfo(freshCtx, jid)
+	groupInfo, err := utils.GetGroupInfoWithRetry(freshCtx, client, jid)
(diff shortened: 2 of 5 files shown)


aldinokemal:
Thanks for addressing the transient disconnect case. Before adding an application-level retry here, could we revisit whether it is still needed on current main?

The pinned Whatsmeow version already handles a disconnected IQ request internally: [`GetGroupInfo`](https://github.com/tulir/whatsmeow/blob/e9a033b2493356741a604a09e50c34b371882338/group.go#L619-L621) goes through [`sendIQ`](https://github.com/tulir/whatsmeow/blob/e9a033b2493356741a604a09e50c34b371882338/request.go#L151-L232), which waits for reconnection (up to five seconds) and retries the interrupted query once. The client also has [automatic reconnect enabled by default](https://github.com/tulir/whatsmeow/blob/e9a033b2493356741a604a09e50c34b371882338/client.go#L270-L292), with the reconnect loop implemented [here](https://github.com/tulir/whatsmeow/blob/e9a033b2493356741a604a09e50c34b371882338/client.go#L577-L650).

This helper therefore mostly adds a second, fixed two-second wait only after the library retry has already failed.

It also does not identify a specific retryable error; any `GetGroupInfo` failure is retried if the client happens to be disconnected when the error is handled. That makes the policy broader and less predictable than the library behavior.

This may help only in the narrow case where reconnection completes just after Whatsmeow has exhausted its five-second wait. If that is the intended scenario, I think we should first measure or reproduce it and decide on a connection-level resilience policy rather than introducing a `GetGroupInfo`-specific retry.

Given the current library behavior, I would avoid merging this helper as-is unless we have evidence that the built-in retry is insufficient.

aldinokemal:
Thanks again for investigating this intermittent disconnect behavior and for the focused implementation. After reviewing the current Whatsmeow behavior, we are going to close this PR rather than merge it: `GetGroupInfo` already benefits from the library’s reconnect-aware IQ retry, and this helper would add a second application-level retry with a separate policy.

The evidence and reasoning are documented in the review comment above. If production data shows that the built-in retry is insufficient, we can reopen the discussion with a reproducible case and a connection-level resilience design. Appreciate the contribution.

Issue #615, opened 2026-03-14
🐞 Send reaction to message doesn't work in group chat


elade:
### What happened?

when trying to send reaction to message in a group chat you get SUCCESS but without reaction actually happening. 

### Reproduction Steps

```bash
1.send post to: /message/{message_id}/reaction
2. message_id - the message you want react to
3. in json post:
   a. phone - the group chat id (instead of private chat id)
   b. emoji - emoji to react with

...
```

### Relevant log output

```bash
{
    "code": "SUCCESS",
    "message": "Reaction sent to 1234567890@g.us (server timestamp: 2026-03-14 14:56:04 +0000 UTC)",
    "results": {
        "message_id": "3EB0123456XX789AB3DE0G",
        "status": "Reaction sent to 1234567890@g.us (server timestamp: 2026-03-14 14:56:04 +0000 UTC)"
    }
}
```

### Application Version

3.8.2

### Verification

- [x] I have checked for existing closed issues and discussions.

aldinokemal:
Fixed in commit 432e974.

**Root cause:** The `ReactMessage` function manually constructed a `MessageKey` without the `Participant` field, which the WhatsApp protocol requires for group chats. Without it, the server accepts the reaction (returning SUCCESS) but silently drops it because it can't resolve which participant's message is being reacted to.

**Fix:** Replaced the manual `MessageKey` construction with whatsmeow's built-in `client.BuildReaction()` helper, which correctly populates the `Participant` field via `BuildMessageKey()` for group chats. The sender JID is resolved from chat storage to identify the original message author.

**Note:** If the original message is not in chat storage (e.g., messages received before chat storage was enabled), the sender cannot be determined and group reactions may still not work. This is a data availability limitation, not a protocol issue.

Issue #535, opened 2026-01-09
🐞 react a message it seems not work


mredodos:
### What happened?

Hi i try to react to a message but it seems not work

### Reproduction Steps

```bash
via n8n or via web ui i sent the react a message but on whatsapp nothing happend
```

### Relevant log output

```bash

[
  
{
  
  
"code": 
"SUCCESS",
  
  
"message": 
"Reaction sent to xxxx@s.whatsapp.net (server timestamp: 2026-01-09 16:53:54 +0000 UTC)",
  
  
"results": 
{
  
  
  
"message_id": 
"xxxx",
  
  
  
"status": 
"Reaction sent to xxxx@s.whatsapp.net (server timestamp: 2026-01-09 16:53:54 +0000 UTC)"
  
  
}
  
}
]
```

### Application Version

8.1.0

### Verification

- [x] I have checked for existing closed issues and discussions.

aldinokemal:
are you using emoji for each?

aldinokemal:
just tested and this is valid, will fix and available on next release

PR #772, opened 2026-07-14
fix: mark group messages as read using the original sender's JID

fra-itc:
## Problem

`MarkAsRead` (src/usecase/message.go) always passes the client's own device JID (`*client.Store.ID`) as the `sender` argument to `whatsmeow.Client.MarkRead(ctx, ids, timestamp, chat, sender, ...)`.

Per whatsmeow's own doc comment on `MarkRead`:

> The second JID parameter (sender) must be set in group chats and must be the user ID who sent the message.

In group chats, whatsmeow only sets the XML `participant` attribute on the receipt node when `sender` is non-empty and `chat.Server` is a group server — always true for a `@g.us` chat. Passing our own device JID there makes the server treat the read receipt as a self-receipt no-op: the *actual* author of the message never gets a blue-tick read confirmation, even though the API call reports success. DMs are unaffected because in a 1:1 chat the chat JID already identifies the sender.

Observed in production: read receipts for group messages from other participants silently did nothing, while DM read receipts worked correctly.

## Fix

Same pattern as #535 (ReactMessage) and #615 (RevokeMessage): resolve the message's original sender from chat storage before calling `MarkRead`.

- Look up the message via `chatStorageRepo.GetMessageByID(messageID)`.
- If found, not `IsFromMe`, and `Sender` parses as a valid JID, use that JID as the `sender` argument to `MarkRead`.
- Otherwise, fall back to the current behavior (`*client.Store.ID`) — correct for DMs and for storage misses.
- Log which path was taken (`sender_source: storage|self-fallback`) for observability.

## Testing

- `go build ./...` passes (verified via `docker build --target builder -f docker/golang.Dockerfile .`).
- No new unit test: like the #535/#615 fixes for the same bug class, the changed code path requires a live whatsmeow client connection (`client.MarkRead` performs real wire I/O), so it isn't unit-testable without a real device — consistent with how #535/#615 shipped.

Fixes the group-chat analogue of the bug fixed for reactions in #535 and #615.


aldinokemal:
Hi, please resolve all AI review comments to complete the PR.

aldinokemal (commented):
I reviewed the complete `MarkAsRead` change against the repository's device-scoped chat storage and the pinned whatsmeow `MarkRead` implementation. Scoping the lookup by device and checking the chat before trusting the stored sender are good improvements.

One major correctness gap remains: every unresolved group-message case still falls back to the client's own JID. In a group, whatsmeow requires the original message author's JID and serializes any non-empty fallback as the `participant` attribute. Passing the current device recreates the original silent no-op while this endpoint can still report success.

I left an inline recommendation to make unresolved group authors an actionable error (or require an explicit validated sender), while retaining fallback behavior only for direct chats. This sender-resolution branch can be unit-tested independently without live WhatsApp I/O.

The PR is currently not mergeable and has no pull-request workflow runs attached to its current head. After addressing the fallback, please rebase and run `gofmt`, `go build ./...`, `go vet ./...`, and `go test ./...`.

src/usecase/message.go
```diff
@@ -47,17 +47,46 @@ func (service serviceMessage) MarkAsRead(ctx context.Context, request domainMess
...
+	// pattern as ReactMessage/RevokeMessage. Messages are keyed by
+	// (id, chat_jid, device_id), so an unscoped GetMessageByID could return a
+	// row from another chat or device with the same id; scope the lookup to
+	// the active device (same helper as send.go's reply-context resolution)
+	// and cross-check the resolved row's ChatJID against the target chat
+	// before trusting its sender. Falls back to the current device JID
+	// (correct for DMs, and whenever no usable sender JID can be resolved).
+	senderJID := *client.Store.ID
+	senderSource := "self-fallback"
+	message, lookupErr := service.chatStorageRepo.GetMessageByIDAndDevice(deviceIDFromContext(ctx), request.MessageID)
```
aldinokemal, line 63: [Major] The self fallback is still invalid for group chats. If storage is missing or stale, the chat check fails, or `Sender` cannot be parsed, this code passes the device's own JID to `MarkRead`. For an `@g.us` target, whatsmeow emits `participant=self`, recreating the existing silent no-op while this endpoint returns success. When the target is a group, fail with an actionable error (or accept and validate an explicit original-sender JID) if the author cannot be resolved; keep this fallback only for direct chats. The resolution logic can be extracted and unit-tested without live wire I/O.

aldinokemal:
Closing this PR because the remaining `MarkAsRead` group-receipt fix has been superseded and merged via #814: https://github.com/aldinokemal/go-whatsapp-web-multidevice/pull/814

#814 carries the focused implementation on current `main`, including chat/device-scoped message lookup, original group sender handling, sender JID validation/normalization, and regression coverage. The unrelated changes that had accumulated in this PR were intentionally not carried over.

Issue #760, opened 2026-07-10
Linking a second device to the same WhatsApp number can silently disconnect or delete the first one


huboperacional:
## Summary

When two device slots are linked to the **same phone number** (WhatsApp allows up to ~4 companion devices per account), the multi-device lifecycle can **delete or hijack the wrong companion's whatsmeow session**, because slot→companion resolution is keyed by the **NonAD JID** (the bare number, `5567...@s.whatsapp.net`) instead of the full **AD JID** (`5567...:32@s.whatsapp.net`, where `:32` identifies the specific companion).

Observed in production: linking a second slot on a number that already had a slot pushed the account past the companion cap; WhatsApp evicted the older companion, and the fork then left the store inconsistent — 6 `whatsmeow_device` rows for one number that maps to 2 live slots, plus a slot that was "alive but unusable" (its client resolved to an invalid session).

## Environment

- `go-whatsapp-web-multidevice` @ `main`, `whatsmeow` current, SQLite store.
- Multi-device mode with named slots (the `devices` table) each mapping to a whatsmeow companion session (`whatsmeow_device`).

## Root cause

The identity written for every slot is the NonAD JID — `src/infrastructure/whatsapp/device_instance.go` (`refreshIdentityLocked`, `NewDeviceInstance`):

```go
d.jid = d.client.Store.ID.ToNonAD().String()   // bare number — loses the :NN companion id
```

Because a slot only remembers the number, three places cannot tell two companions of the same number apart:

1. **`findStoreDeviceByJID`** (`device_manager.go`) resolves a store device by scanning `GetAllDevices()` and returning the **first** row whose `ToNonAD()` matches → a slot's `EnsureClient` can adopt a *different* slot's live session (**session hijack**).

2. **`loadFromRegistry`** (`device_manager.go`): on every boot it deduplicates the `devices` registry by NonAD JID and **`DeleteDeviceRecord`s** the duplicate — silently deleting a legitimate second slot on the same number (**data loss**). It also matches an in-memory instance purely by NonAD, collapsing two companions into one slot.

3. **`LoadExistingDevices`** adopts unreferenced store rows as in-memory instances and the auto-connect loop dials them → endless `websocket close 1005` reconnect churn against dead sessions.

[...]

PR #778, opened 2026-07-21
feat(device): guard bare-number store-row deletes by slot ownership + prune-devices CLI

huboperacional:
## What

Two small protections for the multi-companion-per-number case (issue #760), layered on top of the AD-JID mapping from #762:

### 1. `deleteStoreRowsForJID`: guard the bare-number path by slot ownership

On the AD-JID path, deletion is already exact (only the matching companion suffix). On the **bare-number** path (legacy / not-yet-backfilled slots) the existing guard refuses only when *2+ store rows* share the number. But a single surviving row is still ambiguous when **more than one device slot** claims the bare number — e.g. a sibling's companion was already evicted, leaving one live row that belongs to the *other* slot. Deleting it on behalf of the slot being logged out destroys the sibling's live session.

This adds an up-front check: on the bare-number path, if more than one slot claims the number, refuse and defer to `prune-devices` (after AD JIDs are backfilled). Fails **closed** if slot ownership can't be resolved. It only engages in the legacy/un-backfilled window; the AD path is untouched.

### 2. `prune-devices`: a maintenance CLI to clean orphan companion rows

Over time the `whatsmeow_device` store accumulates companion rows that no slot references any more (re-pairings bump the companion index). This subcommand lists them and, with `--apply`, deletes them. Dry-run by default; skips any number that still has an un-backfilled slot (can't prove such a row is an orphan vs. that slot's live companion). Adds an exported `GetStoreContainers` accessor so the command can read the store without starting the server or dialing WhatsApp.

## Tests

`device_manager_companion_test.go`: multi-slot refusal, single-slot delete proceeds, and fail-closed on a storage error. All existing device-manager/logout tests still pass.

## Verification

`gofmt`, `go build ./...`, `go vet ./...`, and `go test ./...` (full suite) all green in CI.


aldinokemal (changes requested):
Thanks for hardening the multi-companion cleanup path. The new bare-number ownership guard is a good fail-closed improvement, but I found two destructive edge cases in `prune-devices` that should be fixed before merge.

### 1. [P1] Pruning must inspect the raw persisted state, not state already mutated by startup reconciliation

This command is registered under the global `cobra.OnInitialize(initEnvConfig, initApp)`. Before `pruneDevices` reads `ListDeviceRecords`, `initApp` has already called `InitWaCLI` and `dm.LoadExistingDevices(ctx)`.

`LoadExistingDevices` is not read-only: it adopts store rows, persists device records, and backfills legacy slots. Its own comment explicitly says that when a legacy bare-number slot has multiple rows, **the first row wins**. This invalidates the pruner's safety model.

For example:

- registry: one legacy slot for number `N`, with an empty `ad_jid`;
- store: `N:28` and `N:32`, where the live row is not knowable from persisted data;
- startup reconciliation picks one row and persists it as the slot's AD JID;
- `prune-devices` now sees that chosen row as referenced and the sibling row as an orphan;
- `--apply` can delete the live sibling even though the original state was ambiguous and should have been protected.

It also means the documented dry-run is not actually read-only, and store-only orphan rows may be adopted/persisted before candidate selection, making them impossible to prune.

Please give this maintenance command a minimal bootstrap that only opens/migrates the databases and snapshots the persisted registry, without creating a client, loading the device manager, adopting rows, or backfilling identities.

### 2. [P1] Exact AD-JID matching can delete the live legacy row from a separate keys store

The candidate logic applies the same exact-string `referenced[dev.ID.String()]` check to both the primary and keys containers. However, the existing keys-store compatibility logic intentionally uses `sameStoreIdentity`: a legacy bare-number keys row is treated as satisfying its AD-JID counterpart.

A valid state can therefore be:

- registry/primary session: `N:28@s.whatsapp.net`;
- separate keys container: legacy `N@s.whatsapp.net` row;
- no un-backfilled registry slot.

`syncKeysDevice` keeps that bare keys row as the active counterpart, but `prune-devices` does not find its exact string in `referenced`, so it marks the live keys row as an orphan and deletes it with `--apply`.

Please match keys-container rows using the same identity semantics as `sameStoreIdentity` (or migrate/replace the legacy keys row before pruning), rather than exact AD-JID equality only.

### Tests requested

Please add command-level tests covering:

1. a legacy bare slot with multiple sibling store rows, proving pruning does not backfill or delete either row;
2. an unreferenced store-only row, proving it remains a candidate instead of being adopted during command bootstrap;
3. a backfilled AD-JID slot with a legacy bare-number row in a separate keys container, proving the keys row is retained.

Once these cases are addressed, the slot-ownership guard itself looks reasonable to merge.

src/cmd/prune_devices.go
```diff
@@ -0,0 +1,230 @@
...
+		devices, err := cr.container.GetAllDevices(ctx)
+		if err != nil {
+			logrus.Fatalf("[PRUNE] failed to enumerate %s store devices: %v", cr.label, err)
+		}
+		for _, dev := range devices {
+			if dev == nil || dev.ID == nil {
+				continue
+			}
+			adJID := dev.ID.String()
+			if referenced[adJID] {
```
aldinokemal, line 143: **Blocker — deletes the live session when `devices` is empty.**

On an install paired before the multi-device registry existed (v7.x upgraded to v8/v9), the `devices` table is empty while a live `whatsmeow_device` row exists. `InitWaCLI` keys its instance by the AD JID (`init.go:127`) and registers it with `EnsureDefault`, which — unlike `AddDevice` — never calls `SaveDeviceRecord`. `event_handler.go:238` then refuses to persist it because `instance.ID()` contains `@`, and `LoadExistingDevices` takes the non-persisting `exactMatch` branch. Nothing ever backfills it.

So `referenced` and `unbackfilledNumbers` are both empty and neither gate fires:

```
Slots referenced (kept) by AD JID: 0
NUMBER                        AD JID                           CONTAINER
6281234567890@s.whatsapp.net  6281234567890:28@s.whatsapp.net  primary
[PRUNE] deleted 6281234567890:28@s.whatsapp.net from primary store
```

`whatsmeow_device` went 1 row → 0. With `--db-keys-uri` set, both containers are wiped. The user is silently unpaired.

This shape is exactly the audience for "prune orphan sessions", so it needs a hard guard: if a store row is the only session and the registry knows nothing about the number, that is evidence the registry is incomplete — not evidence the row is an orphan.

aldinokemal (commented):
Thanks for this — the #760 problem is real and the AD-JID direction is right. I went through both halves carefully and I can't take it as-is: I was able to reproduce several paths where `prune-devices --apply` deletes a **live** session rather than an orphan, and the new guard in `deleteStoreRowsForJID` changes logout semantics in ways that leave rows behind or resurrect sessions.

Everything below was reproduced against seeded SQLite fixtures and a binary built from this branch — not just read off the diff.

### The core issue with `prune-devices`

It decides what is an orphan from the `devices` registry table, but that table is not a reliable inverse of the whatsmeow store:

1. A default/legacy install has **zero** rows there — `EnsureDefault` never calls `SaveDeviceRecord`, and `event_handler.go:238` refuses to persist an instance whose ID contains `@`. So the only live session is listed as an orphan.
2. The command's own `initApp` bootstrap (`LoadExistingDevices`) **rewrites that table before the scan reads it** — adopting orphan rows into slots, and backfilling legacy slots with the *oldest* store row, which disarms safety gate 1 and points it at the live companion.
3. `config.ChatStorageURI` is hard-coded CWD-relative with no flag or env binding, while the store half honours `--db-uri` — so the two halves can describe different installations.

Because of (2), the dry-run is also not read-only: it writes to all three databases.

### On shape

Rather than a second process that must re-run the entire app bootstrap just to read two tables, this reconciliation looks better placed inside the service (a maintenance endpoint, or a startup pass) where the registry and the containers are already open and consistent.

Also worth noting: the guard and the prune scan ask the same question — *which store rows does this slot own?* — twice, in two subtly different ways. One shared resolver returning the owned rows would let `deleteStoreRowsForJID` delete **exactly** instead of refusing, and would give the prune scan the same semantics for free.

15 inline comments below, roughly most-severe first.

src/cmd/prune_devices.go
```diff
@@ -0,0 +1,230 @@
...
+	// deliberately do NOT build a client or connect: this is a maintenance path.
+	primary, keys := whatsapp.GetStoreContainers()
+	if primary == nil {
+		logrus.Fatalf("[PRUNE] whatsmeow store container not initialized")
+	}
+	if chatStorageRepo == nil {
+		logrus.Fatalf("[PRUNE] chat storage repository not initialized")
+	}
+
+	records, err := chatStorageRepo.ListDeviceRecords()
```
aldinokemal, line 100: **Blocker — the registry half can't be pointed at the same install as `--db-uri`.**

`chatStorageRepo` resolves from `config.ChatStorageURI`, which is the literal `"file:storages/chatstorage.db"` (`config/settings.go:76`) with **no persistent flag and no viper/env binding** — it is only read at `cmd/root.go:622`. The store half, meanwhile, honours `--db-uri` / `DB_URI`. `initApp` silently creates the chatstorage file when absent, so a mismatch produces no warning at all.

Verified:

```
dir A (real install):  prune-devices                          -> referenced: 1, candidates: 0
dir B (any other cwd): prune-devices --db-uri file:A/whatsapp.db
                       -> referenced: 0, candidates: 1
                       --apply --yes -> "[PRUNE] deleted 628111:14@s.whatsapp.net"   (A is now empty, exit 0)
```

Two realistic triggers: the documented local workflow runs from `src/` so the registry lives in `src/storages/`; and for Postgres deployments it is natural to assume no volume is needed — `docker run --rm -e DB_URI=postgres://… gowa prune-devices --apply --yes` gets an ephemeral empty registry and deletes every session in the shared database.

At minimum this command should refuse to run when the registry it opened was newly created, and `ChatStorageURI` should be flag/env-bindable.

src/cmd/prune_devices.go
```diff
@@ -0,0 +1,230 @@
...
+				continue
+			}
+			adJID := dev.ID.String()
+			if referenced[adJID] {
+				continue // referenced by a backfilled slot -> live, keep it
+			}
+			nonAD := dev.ID.ToNonAD().String()
+
+			// Safety gate 1: protect rows whose number still has an un-backfilled slot.
+			if unbackfilledNumbers[nonAD] {
```
aldinokemal, line 149: **Blocker — safety gate 1 is disarmed by this command's own bootstrap, and it selects the live row.**

`cobra.OnInitialize` runs `initApp` before `Run` (cobra `command.go:959`), so `LoadExistingDevices` has already backfilled every un-backfilled slot that has a matching store row (`device_manager.go:584-588`) by the time `unbackfilledNumbers` is built. Those slots are no longer un-backfilled, so the gate never engages for them.

Worse, the row it picks is not arbitrary. `getAllDevicesQuery` has no `ORDER BY`, so on SQLite the first row is deterministically the lowest rowid — the **oldest** companion. After a re-pair the oldest row is the dead one:

```
[DEVICE_MANAGER] backfilling companion 6284444444444:10@s.whatsapp.net for device slot-legacy
Numbers protected by un-backfilled slots: 0          <- gate 1 defeated
NUMBER                        AD JID                           CONTAINER
6284444444444@s.whatsapp.net  6284444444444:11@s.whatsapp.net  primary   <- the LIVE one

--apply --yes:
[PRUNE] deleted 6284444444444:11@s.whatsapp.net from primary store
[PRUNE] deleted 6284444444444:11@s.whatsapp.net from keys store
store AFTER: only 6284444444444:10@... (the dead companion) remains
```

So the deterministic SQLite outcome is "keep the dead session, delete the live one". On Postgres seq-scan order makes it a coin flip.

The gate can only work if the record snapshot is taken **before** `LoadExistingDevices` runs — or if this doesn't ride `initApp` at all.

src/cmd/prune_devices.go
```diff
@@ -0,0 +1,230 @@
...
+	// Referenced set = every non-empty persisted AD JID. Un-backfilled slots (empty
+	// AD JID) instead contribute their bare number to the protected-numbers set
+	// (safety gate 1).
+	referenced := make(map[string]bool)
+	unbackfilledNumbers := make(map[string]bool)
+	for _, rec := range records {
+		if rec == nil {
+			continue
+		}
+		if adJID := strings.TrimSpace(rec.ADJID); adJID != "" {
```
aldinokemal, line 114: **Blocker — a stale `ad_jid` strips all protection from its number.**

This `else if` means a record with a non-empty `ad_jid` contributes to `referenced` only, never to `unbackfilledNumbers`. If that `ad_jid` is stale, it protects nothing (it matches no row), and the slot's *actual* live row is unguarded:

```
Slots referenced (kept) by AD JID: 1
Numbers protected by un-backfilled slots: 0
[PRUNE] deleted 6286666666666:11@s.whatsapp.net from primary store
[PRUNE] deleted 6286666666666:11@s.whatsapp.net from keys store
store AFTER: (empty)
```

Stale `ad_jid` is reachable and permanent for auto-created slots (`device_id` == the bare JID, created at `device_manager.go:608`): `event_handler.go:238` skips persistence when the ID contains `@`, and I checked all five `SaveDeviceRecord` call sites — none refreshes a non-empty stale `ad_jid`. Entry points include `companionClaimedByOther` forcing a fresh pairing, whatsmeow's own `cli.Store.Delete` on a 401, and a previous `prune --apply` (i.e. it is self-perpetuating).

The missing signal is right here: nothing checks that a `referenced` AD JID actually corresponds to a live store row. A referenced-but-absent AD JID means the registry is stale and the number should be protected, not exposed.

src/cmd/prune_devices.go
```diff
@@ -0,0 +1,230 @@
...
+	for _, cr := range containers {
+		devices, err := cr.container.GetAllDevices(ctx)
+		if err != nil {
+			logrus.Fatalf("[PRUNE] failed to enumerate %s store devices: %v", cr.label, err)
+		}
+		for _, dev := range devices {
+			if dev == nil || dev.ID == nil {
+				continue
+			}
+			adJID := dev.ID.String()
```
aldinokemal, line 142: **Blocker — legacy bare-number rows (Device == 0) are unconditional candidates.**

A store row whose JID has no device suffix can never land in `referenced` (which is keyed on `ad_jid`), and its number won't be in `unbackfilledNumbers` if the slot sharing it is backfilled. So it is always a candidate — even with a completely correct registry:

```
Slots referenced (kept) by AD JID: 1
NUMBER                 AD JID                 CONTAINER
628111@s.whatsapp.net  628111@s.whatsapp.net  primary     <- listed for deletion
```

Every other place in the codebase treats this row class as possibly-live and refuses to touch it — `deleteStoreRowsForJID` at `device_manager.go:271-275` ("may be a legacy slot's usable session and must never be taken along"), its dedicated test `TestDeleteStoreRowsForJID_ADTargetLeavesBareNumberRowAlone`, `sameStoreIdentity`, and `syncKeysDevice`'s "Legacy bare-number rows still match their AD counterpart". This scan drops that guard.

In a `DB_KEYS_URI` deployment it's worse than losing an orphan: the keys container legitimately holds the live device under the legacy bare form (precisely what `syncKeysDevice` tolerates), so deleting it cascades away every FK-linked key table while the primary row survives.

Note also that the candidate table renders NUMBER and AD JID identically for these rows, so an operator reviewing the dry-run has no way to tell it apart from a real orphan.

src/infrastructure/whatsapp/device_manager.go
```diff
@@ -213,6 +213,41 @@ func (m *DeviceManager) deleteStoreRowsForJID(ctx context.Context, jid string) e
...
+			// is worse than useless. If we cannot enumerate slots we cannot prove this
+			// bare-number delete is unambiguous, so refuse rather than risk a sibling's
+			// live session. The caller surfaces the error; a retry (or prune-devices)
+			// cleans up later.
+			logrus.WithError(lerr).Errorf("[DEVICE_MANAGER] cannot resolve slot ownership for number %s; refusing to delete store rows", bare)
+			return lerr
+		}
+		slotsClaiming := 0
+		for _, rec := range records {
+			if rec != nil && rec.JID == bare {
```
aldinokemal, line 239: **The guard misses the exact case its own comment describes.**

Ownership is inferred from `rec.JID`, but a slot that is mid-pairing — or that was just keep-slot logged out — has an empty `rec.JID` while owning a live store row. It is therefore invisible to this count.

Records `{slot-a: jid=628…, ad_jid: ""}` and `{slot-b: jid: "", ad_jid: ""}`, one live row `628…:55` owned by slot-b. Logging out slot-a gives `slotsClaiming == 1`, the guard passes, `deleteFrom` finds exactly one match, and slot-b's live session is deleted. Verified: `surviving rows = 0`; the control with slot-b's `rec.JID` populated is correctly protected.

That is verbatim the comment above — *"a sibling's companion was evicted, leaving one live row that belongs to the OTHER slot"*.

The window is normal operation, not a crash: whatsmeow saves the store row at `pair.go:216` *before* dispatching `PairSuccess`, `handlePairSuccess` persists nothing to the registry, and the record only gains its JID at the later `Connected` event (`event_handler.go:234-250`) — seconds normally, indefinitely if the reconnect never completes. `resetDeviceKeepSlot` writes `JID: ""` too.

`DeleteDevice` cascades across all the FK-linked key tables, so the freshly paired session is unrecoverable.

src/cmd/prune_devices.go
```diff
@@ -0,0 +1,230 @@
...
+			fmt.Println("Aborted. Nothing deleted.")
+			return
+		}
+	}
+
+	deleted := 0
+	for _, c := range candidates {
+		if err := c.container.DeleteDevice(ctx, c.dev); err != nil {
+			logrus.WithError(err).Errorf("[PRUNE] failed to delete %s from %s store", c.adJID, c.label)
+			continue
```
aldinokemal, line 210: **`continue` here can half-delete a primary/keys pair and cascade away all key material.**

If the primary delete fails and the keys delete succeeds, the device keeps its primary row (so `LoadExistingDevices` re-adopts it and it appears paired and connects) but has lost every Signal session, identity, pre-key and sender key:

```
[PRUNE] failed to delete ...:28@s.whatsapp.net from primary store  error="database is locked"
[PRUNE] deleted ...:28@s.whatsapp.net from keys store
  candidates: 2 / deleted: 1
primary device/sessions/identities: 1/1/1     <- still paired
keys    device/sessions/identities: 0/0/0     <- all Signal material gone
```

Exit code is **0**, so a wrapper script sees success.

All the premises hold: the whatsmeow schema declares `FOREIGN KEY (our_jid) REFERENCES whatsmeow_device(jid) ON DELETE CASCADE` on 12+ tables, `InitWaDB` (`database.go:35`) sets `_foreign_keys=on` on the keys DB as well, and `key_cache.go` points Identities/Sessions/PreKeys/SenderKeys/MsgSecrets at the keys container.

A candidate should be deleted from both containers or neither, and a partial failure should be a non-zero exit.

src/infrastructure/whatsapp/device_manager.go
```diff
@@ -213,6 +213,41 @@ func (m *DeviceManager) deleteStoreRowsForJID(ctx context.Context, jid string) e
...
+		bare := target.ToNonAD().String()
+		records, lerr := m.storage.ListDeviceRecords()
+		if lerr != nil {
+			// Fail closed: a safety guard that silently disables itself on a storage error
+			// is worse than useless. If we cannot enumerate slots we cannot prove this
+			// bare-number delete is unambiguous, so refuse rather than risk a sibling's
+			// live session. The caller surfaces the error; a retry (or prune-devices)
+			// cleans up later.
+			logrus.WithError(lerr).Errorf("[DEVICE_MANAGER] cannot resolve slot ownership for number %s; refusing to delete store rows", bare)
+			return lerr
```
aldinokemal, line 235: **Fail-closed here, but the steps that destroy the identity still run.**

`deleteStoreRowsForJID` is now the only step that can fail — and it runs *before* the two steps that destroy its own inputs: `resetDeviceKeepSlot` (blanks `jid`/`ad_jid`) and `RemoveDevice` (deletes the record). Both callers run them unconditionally:

```go
// keepSlotLogout :432-433
firstErr = errors.Join(firstErr, m.deleteStoreRowsForJID(ctx, jid))
firstErr = errors.Join(firstErr, m.resetDeviceKeepSlot(deviceID))

// PurgeDevice :359-362
recordErr(m.deleteStoreRowsForJID(ctx, jid))
m.RemoveDevice(deviceID)
```

Verified on `keepSlotLogout`: attempt 1 returns `storage unavailable` with both rows intact and the slot blanked in memory *and* on disk; attempt 2 with healthy storage returns `<nil>` and still deletes nothing, because `storeIdentity(inst)` is now `""` and the function short-circuits at `:208`.

Verified on `PurgeDevice`: returns the error (HTTP 500 via `usecase/device.go:94-97`) while `RemoveDevice` has already deleted the record — `slot still in registry: false`, `registry: 0 records`, both rows orphaned. The retry the 500 invites returns 200 OK and cleans nothing, and on restart `LoadExistingDevices` re-creates the "deleted" device under the bare-JID id, with its chat data already destroyed by `DeleteDeviceData`.

Both HTTP callers (`usecase/app.go:269-272`, `usecase/device.go:123-125`) also return before the `DEVICE_LOGGED_OUT` websocket broadcast, so the UI keeps showing the device as connected.

So the comment just above — *"a retry (or prune-devices) cleans up later"* — isn't right for the retry half: the retry's precondition was destroyed by the same call that failed. Gating the identity-erasing steps on the delete having actually happened would fix this and make the comment true.

src/infrastructure/whatsapp/device_manager.go
```diff
@@ -213,6 +213,41 @@ func (m *DeviceManager) deleteStoreRowsForJID(ctx context.Context, jid string) e
...
+		for _, rec := range records {
+			if rec != nil && rec.JID == bare {
+				slotsClaiming++
+			}
+		}
+		if slotsClaiming > 1 {
+			// A safe skip (not an execution error): mirror the store-row-count guard below,
+			// which also warns and returns without deleting.
+			logrus.Warnf("[DEVICE_MANAGER] %d slots claim number %s and no AD JID is known; refusing to delete any whatsmeow rows -- run prune-devices after the AD JIDs are backfilled", slotsClaiming, bare)
+			return nil
```
aldinokemal, line 247: **Silent skip: a restart grafts the dead session onto a sibling slot.**

Returning `nil` reports a clean logout, but the row survives while `resetDeviceKeepSlot` still blanks the slot. On the next boot the sibling still carries `JID = nonAD`, so `legacyMatch` wins and the row is backfilled onto **the sibling — not the slot that was logged out**:

```
[DEVICE_MANAGER] 2 slots claim number 6280000001001@... ; refusing to delete...
keepSlotLogout(slot-a) err = <nil>          <- reported as success
--- RESTART ---
[DEVICE_MANAGER] backfilling companion 6280000001001:40@s.whatsapp.net for device slot-b
```

slot-b now owns the session slot-a just unlinked on the phone. `SetAutoConnectAfterBooting` dials it, and the resulting `LoggedOut` runs `keepSlotLogout` against slot-b — wiping the identity of a slot the user never asked to log out.

If the skip is a real outcome the caller needs to know about, it probably shouldn't be indistinguishable from a successful delete.

src/cmd/prune_devices.go
```diff
@@ -0,0 +1,230 @@
...
+	dev       *store.Device
+	adJID     string
+	nonAD     string
+}
+
+func pruneDevices(_ *cobra.Command, _ []string) {
+	ctx := context.Background()
+
+	// Reuse the containers + registry constructed by initApp (cobra.OnInitialize). We
+	// deliberately do NOT build a client or connect: this is a maintenance path.
```
aldinokemal, line 91: **This comment is the crux of the problem: `initApp` doesn't just build containers, it rewrites the registry.**

`LoadExistingDevices` runs inside `initApp` (`root.go:680`) and adopts unreferenced store rows before the scan ever reads records. Seeded with 3 genuine orphans plus one empty named slot, a plain **dry-run** reported 1 candidate, and the `devices` table afterwards showed:

```
[DEVICE_MANAGER] matching orphaned device empty-slot with JID 6282222222222@s.whatsapp.net

empty-slot | 6282222222222@s.whatsapp.net | 6282222222222:20@s.whatsapp.net   <- MUTATED
6283333333333@s.whatsapp.net | 6283333333333@... | 6283333333333:30@...       <- INSERTED
```

Only rows hitting the `numberClaimed` branch (`device_manager.go:594-597`) survive into the candidate list; rows adopted via `legacyMatch`, `orphanSlot`, or the `AddDevice` create branch are all swallowed. So the command mostly cannot see the orphans it exists to find.

And `applyStoreJID` also swaps the victim slot's chat-storage partition key — a user's freshly created, not-yet-paired slot gets silently rebound to an unrelated dead companion, by the mode documented as read-only.

src/cmd/prune_devices.go
```diff
@@ -0,0 +1,230 @@
...
+     backfilled), it claims only a bare number. Any candidate row carrying such a number
+     is SKIPPED, because we cannot prove it is an orphan rather than that slot's live
+     companion. Backfill the AD JIDs (reconnect the slots) before pruning those numbers.
+  2. No per-row age guard is available: whatsmeow's store.Device exposes no creation
+     timestamp, so a freshly-created (mid-pairing) row cannot be detected here.
+
+*** STOP THE GOWA SERVICE BEFORE RUNNING --apply. ***
+This command opens the same SQLite session DB the running service uses. Deleting a row
+while the service is live risks concurrent-access corruption and can destroy a session
+that is being paired at that moment. Dry-run (the default) is read-only and safe to run
```
aldinokemal, line 54: **This sentence isn't accurate — the dry-run writes to all three databases.**

Measured on a bare `prune-devices` with zero candidates:

- `chatstorage.db` sha256 changed and `devices.updated_at` moved from the seeded `2026-01-01` to the run time (the registry writes from `LoadExistingDevices`).
- `whatsapp-keys.db` went from 0 to 1 `whatsmeow_device` rows — `root.go:675` `InitWaCLI` → `init.go:65` `keysDB.PutDevice`, copying noise key, identity key, signed pre-key and adv keys into a second DB.
- Run in an empty directory, the dry-run created all three DB files plus `-wal`/`-shm` from nothing (`root.go:667` `InitializeSchema`, `root.go:669/672` `InitWaDB` → `container.Upgrade`).

So an operator who trusts this line and runs the dry-run against a live service gets exactly the concurrent-SQLite-write hazard the paragraph above reserves for `--apply`.

src/infrastructure/whatsapp/device_manager.go
```diff
@@ -213,6 +213,41 @@ func (m *DeviceManager) deleteStoreRowsForJID(ctx context.Context, jid string) e
...
+			logrus.WithError(lerr).Errorf("[DEVICE_MANAGER] cannot resolve slot ownership for number %s; refusing to delete store rows", bare)
+			return lerr
+		}
+		slotsClaiming := 0
+		for _, rec := range records {
+			if rec != nil && rec.JID == bare {
+				slotsClaiming++
+			}
+		}
+		if slotsClaiming > 1 {
```
aldinokemal, line 243: **This counts records that are never loaded as slots, which can deadlock cleanup permanently.**

`loadFromRegistry` deliberately *keeps* records it refuses to load: auto-created records shadowed by a named slot (`:673-676`) and duplicate-identity records (`:683-687`, "record kept; remove the slot explicitly if stale"). Those still carry `JID == bare` and inflate `slotsClaiming`.

With one real live slot plus one such phantom: `live slots=1, registry records=2`, and the guard logs `2 slots claim number …003@s.whatsapp.net … refusing to delete any whatsmeow rows`. The remedy the warning suggests doesn't help either — the phantom's empty `ad_jid` puts the number in `unbackfilledNumbers`, so `prune-devices` skips the row too (`prune_devices.go:116,149`). And the phantom can't be removed through the API: it isn't in `m.devices`, so `ResolveDevice` returns "device not found".

Marking this one as *plausible* rather than confirmed — a normal boot usually backfills the live slot, which routes logout to the AD path where this guard never runs. It needs the slot to stay un-backfilled: `m.store.GetAllDevices` failing at boot (the error is discarded by `_ =` at `root.go:680`), or the row living only in the keys container, which `LoadExistingDevices` never scans.

Related, and cheap to fix: the loop above never consults `rec.ADJID`, so a fully backfilled sibling — whose AD JID provably can't be the surviving row — still blocks the delete, and the log line "and no AD JID is known" is then simply false. Subtracting rows owned by a sibling's recorded `ADJID` would let this delete exactly instead of refusing.

src/cmd/prune_devices.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	// --apply path.
+	if !pruneDevicesYes {
+		fmt.Printf("\nAbout to DELETE %d orphan companion row(s).\n", len(candidates))
+		fmt.Print("The GOWA service MUST be stopped. Type 'y' to confirm: ")
+		reader := bufio.NewReader(os.Stdin)
+		line, _ := reader.ReadString('\n')
+		ans := strings.ToLower(strings.TrimSpace(line))
+		if ans != "y" && ans != "yes" {
```
aldinokemal, line 200: **Non-interactive `--apply` silently no-ops and exits 0.**

The read error is discarded, so a closed or non-TTY stdin yields `ans == ""` — neither `y` nor `yes` — and this returns. Measured exit code **0**, no `Summary:` block (this is the only exit that omits it), and no hint that `--yes` was required.

Docker/compose/cron is the primary deployment shape here, so `docker compose exec -T gowa /app/whatsapp prune-devices --apply` prints an abort line into a log nobody reads and reports success. An operator scripting the documented "stop the service, then prune" procedure has no programmatic way to tell a clean no-op from a completed prune.

Failing closed is right; exiting 0 isn't.

src/cmd/prune_devices.go
```diff
@@ -0,0 +1,230 @@
...
+		}
+	}
+
+	// Enumerate both containers (keys only when it is a distinct container).
+	type containerRef struct {
+		label     string
+		container *sqlstore.Container
+	}
+	containers := []containerRef{{"primary", primary}}
+	if keys != nil && keys != primary {
```
aldinokemal, line 127: **Counts are per store row rather than per session, and this is a pointer comparison.**

In the normal `DB_KEYS_URI` deployment with genuinely distinct files, 2 logical orphan sessions are reported as `candidates: 4` / `would delete 4 orphan companion row(s)`, because each session has a keys mirror.

Separately, `initApp` calls `InitWaDB` once per URI (`root.go:669-672`), so setting `DB_KEYS_URI` equal to `DB_URI` produces two distinct pointers over one file and this check never fires. Verified: the same row listed twice (primary and keys), `candidates: 2` for one physical row, and under `--apply` the second `DELETE` matched zero rows yet still incremented `deleted` — whatsmeow's `DeleteDevice` never inspects `RowsAffected`.

Two smaller ones in the same output: `Slots referenced (kept) by AD JID` counts distinct non-empty `ad_jid` *values* (with 3 slots — two sharing an ad_jid, one stale with no store row — it printed 2 while only 1 row per container was kept), and the dry-run tally prints twice, once at the `[dry-run] would delete N` line and again inside `printPruneSummary`.

src/cmd/prune_devices.go
```diff
@@ -0,0 +1,230 @@
...
+// pruneCandidate is one whatsmeow store row considered for deletion.
+type pruneCandidate struct {
+	container *sqlstore.Container
+	label     string
+	dev       *store.Device
+	adJID     string
+	nonAD     string
+}
+
+func pruneDevices(_ *cobra.Command, _ []string) {
```
aldinokemal, line 87: **No tests for this file.**

The classification logic that decides which sessions get deleted — `referenced`/`unbackfilledNumbers` construction and safety gate 1 — is pure and deterministic, but it is inlined here alongside stdin/stdout I/O, so it can't be exercised without running the binary. The `cmd` package is already testable (`src/cmd/rest_test.go` exists).

Every data-loss case above was reproducible by seeding two tables and calling that logic. A single table-driven test over `(records, store rows) -> (candidates, skipped)` would have caught the empty-registry case, the stale-`ad_jid` case and the bare-number-row case before review.

The PR description cites `device_manager_companion_test.go` as covering this, but those three tests only exercise `deleteStoreRowsForJID`.

One related note on the guard's tests: `keepSlotStubStorage.ListDeviceRecords` returns `nil, nil`, so every pre-existing logout/purge test now runs the new guard with `slotsClaiming == 0`. None of them would catch a regression in it — worth giving the base stub a realistic record set.

PR #784, opened 2026-08-01
feat(send): add calendar event messages (/send/event + whatsapp_send_event MCP tool)

kurotsuke:
> [!IMPORTANT]
> **Depends on tulir/whatsmeow#1214** — event messages must be framed `type="event"` with a `<meta event_type="creation"/>` node or recipient clients silently drop them. Verified end-to-end (renders with working RSVP) on a build with that patch; once it lands upstream, a routine whatsmeow bump here makes this PR fully functional.

## What

Adds WhatsApp **calendar event** sending — the native event message type with RSVP (Attending / Not attending), as seen in group and 1:1 chats.

- `POST /send/event` — REST endpoint
- `whatsapp_send_event` — MCP tool
- OpenAPI spec + readme updated

## Example

```bash
curl -X POST http://localhost:3000/send/event \
  -H 'Content-Type: application/json' \
  -d '{
    "phone": "6289685024421@s.whatsapp.net",
    "name": "Team Meeting",
    "description": "Quarterly planning session",
    "start_time": 2063137000,
    "end_time": 2063140600,
    "location_name": "Head Office",
    "extra_guests_allowed": false
  }'
```

`name` and `start_time` (unix seconds) are required; `description`, `end_time`, `location_name`, `extra_guests_allowed`, and the standard `duration` (disappearing messages) are optional.

## How

Mirrors the existing `SendPoll` flow layer by layer (domain DTO → ozzo validation → usecase → REST/MCP):

- Builds `waE2E.EventMessage` and sends it through the existing `wrapSendMessage` path, so chat storage and error normalization behave like every other send.
- The creation message carries a random 32-byte `MessageSecret` in `MessageContextInfo` — same mechanism as poll creation. This is required: recipients encrypt their RSVP responses with the event's message secret, so without it the event renders but nobody can respond.
- `duration` maps to `ContextInfo.Expiration` like the other message types.
- Location is name-only (`LocationMessage.Name`); coordinates are omitted.

## Validation

- `name`, `start_time`, `phone` required; international phone format check
- `end_time`, when set, must be after `start_time`
- `duration` restricted to the existing disappearing-message whitelist

## Tests

- Table-driven tests for `ValidateSendEvent` (6 cases) in `validations/send_validation_test.go`
- `go build ./...`, `go vet ./...`, and the full `go test ./...` suite pass (11 packages ok)

### DIFF
--- a/src/ui/mcp/send.go
+++ b/src/ui/mcp/send.go
@@ -786,14 +786,19 @@ func (s *SendHandler) handleSendEvent(ctx context.Context, request mcp.CallToolR
 		eventRequest.EndTime = &endTime
 	}
 
-	if args := request.GetArguments(); args != nil {
-		if _, ok := args["duration"]; ok {
-			duration, err := request.RequireInt("duration")
-			if err != nil {
-				return nil, err
-			}
-			eventRequest.Duration = &duration
+	// duration goes through the same strict integer path as the timestamps:
+	// RequireInt casts float64 to int, so fractional values like 86400.5
+	// would silently truncate onto the duration whitelist.
+	duration64, ok, err := int64Arg(request, "duration")
+	if err != nil {
+		return nil, err
+	}
+	if ok {
+		if duration64 < math.MinInt32 || duration64 > math.MaxInt32 {
+			return nil, errors.New("argument \"duration\" is out of range")
 		}
+		duration := int(duration64)
+		eventRequest.Duration = &duration
 	}
 
 	res, err := s.sendService.SendEvent(ctx, eventRequest)
--- a/src/ui/mcp/send.go
+++ b/src/ui/mcp/send.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strconv"
 
 	domainSend "github.com/aldinokemal/go-whatsapp-web-multidevice/domains/send"
 	mcpHelpers "github.com/aldinokemal/go-whatsapp-web-multidevice/ui/mcp/helpers"
@@ -697,7 +698,40 @@ func (s *SendHandler) toolSendEvent() mcp.Tool {
 		mcp.WithBoolean("extra_guests_allowed",
 			mcp.Description("Whether attendees may bring extra guests (default: false)"),
 		),
+		mcp.WithNumber("duration",
+			mcp.Description("Optional disappearing message duration in seconds (0, 86400, 604800, 7776000)"),
+		),
 	)
+}
+
+// int64Arg reads an MCP number argument as int64 without routing through the
+// platform-sized int helpers, which truncate unix timestamps past 2038 on
+// 32-bit builds (e.g. the armv7 image).
+func int64Arg(request mcp.CallToolRequest, key string) (int64, bool, error) {
+	args := request.GetArguments()
+	if args == nil {
+		return 0, false, nil
+	}
+	val, exists := args[key]
+	if !exists || val == nil {
+		return 0, false, nil
+	}
+	switch v := val.(type) {
+	case float64:
+		return int64(v), true, nil
+	case int64:
+		return v, true, nil
+	case int:
+		return int64(v), true, nil
+	case string:
+		parsed, err := strconv.ParseInt(v, 10, 64)
+		if err != nil {
+			return 0, false, fmt.Errorf("argument %q cannot be converted to int64", key)
+		}
+		return parsed, true, nil
+	default:
+		return 0, false, fmt.Errorf("argument %q is not an integer", key)
+	}
 }
 
 func (s *SendHandler) handleSendEvent(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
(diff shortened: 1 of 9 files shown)


kurotsuke:
Addressed all review feedback in 00fe091:

- **int64 timestamps (Codex):** `start_time`/`end_time` are now parsed with a local `int64Arg` helper reading the raw MCP arguments instead of `RequireInt`/`GetInt`, which route through platform-sized `int` and would truncate post-2038 timestamps on the 32-bit armv7 build.
- **Explicit invalid `end_time` (Codex):** a supplied `end_time` is now always passed through to validation (presence-checked, not value-checked), so zero/negative values are rejected with `end_time must be after start_time` instead of being silently dropped.
- **`duration` parity (CodeRabbit):** `whatsapp_send_event` now exposes the optional disappearing-message `duration`, wired with the same presence-check pattern as `whatsapp_forward_message`.
- **Extra test cases (CodeRabbit nitpick):** added negative `start_time` and `end_time == start_time` cases to `TestValidateSendEvent`.

`go build ./...`, `go vet ./...`, `gofmt`, and the full test suite pass.

kurotsuke:
Addressed in the latest commit: `int64Arg` now rejects NaN, infinities, fractional values (`math.Trunc(v) != v`), and floats outside the int64 range instead of silently truncating them. Build, vet, gofmt, and the full test suite pass.

kurotsuke:
Two updates after end-to-end testing on live accounts:

1. **Added explicit lifecycle/reminder fields** (`isCanceled`, `isScheduleCall`, `hasReminder`, `reminderOffsetSec`) to the built `EventMessage` — captured a phone-created event via linked-device sync and recipients ignore event creations without them.
2. **Dependency note:** rendering also requires a small whatsmeow fix — event messages must be framed `type="event"` with a `<meta event_type="creation"/>` node (mirroring the existing poll handling). Submitted upstream as tulir/whatsmeow#1214. Once that lands and this repo bumps whatsmeow, this PR is fully functional — verified end-to-end (event renders with working RSVP on the recipient phone) using a build with that patch applied.

src/usecase/send.go
```diff
@@ -1354,6 +1355,74 @@ func (service serviceSend) SendPoll(ctx context.Context, request domainSend.Poll
...
+	}
+
+	msg := &waE2E.Message{
+		EventMessage:       eventMessage,
+		MessageContextInfo: &waE2E.MessageContextInfo{MessageSecret: msgSecret},
+	}
+
+	content := "📅 " + request.Name
+
+	ts, err := service.wrapSendMessage(ctx, client, dataWaRecipient, msg, content)
```
aldinokemal, line 1416: **Blocker — the current dependency makes this success path misleading.** This branch still pins whatsmeow at `662ad1dc6900`, which does not frame `EventMessage` stanzas as `type="event"` with `<meta event_type="creation"/>`. Upstream PR `tulir/whatsmeow#1214` that adds that behavior is still open. As a result, this call can return a valid message ID/server timestamp while recipient clients silently drop the event. Please bump to a whatsmeow revision containing #1214 or carry the equivalent patch before exposing this endpoint as functional.

src/ui/mcp/send.go
```diff
@@ -665,6 +668,142 @@ func (s *SendHandler) handleSendPoll(ctx context.Context, request mcp.CallToolRe
...
+	if err != nil {
+		return nil, err
+	}
+	if ok {
+		eventRequest.EndTime = &endTime
+	}
+
+	if args := request.GetArguments(); args != nil {
+		if _, ok := args["duration"]; ok {
+			duration, err := request.RequireInt("duration")
```
aldinokemal: **Major — reject fractional duration instead of truncating it.** `duration` is declared with `WithNumber`, so MCP JSON supplies it as `float64`. In mcp-go v0.56.0, `RequireInt` performs `int(v)` without checking the fractional component; for example, `86400.5` becomes `86400` and then passes `validateDuration`. Parse this with the same strict checked-integer path used for `start_time`/`end_time` (then bounds-check before converting to `int`) so malformed values are rejected rather than silently changing semantics.

aldinokemal (commented):
Reviewed the REST and MCP entry points, validation, protobuf construction, disappearing-message context, message-secret generation, whatsmeow integration, existing review threads, dependency pin, and available status checks.

There is one merge blocker: this branch still pins `go.mau.fi/whatsmeow` at commit `662ad1dc6900`, while the required event stanza framing change is still only in open upstream PR `tulir/whatsmeow#1214`. With the currently pinned dependency, `SendMessage` frames the payload as a normal text message; WhatsApp may accept it and return a server timestamp, but recipient clients silently discard the event. That means both `/send/event` and `whatsapp_send_event` can report success for a message that is not delivered visibly. Please include a whatsmeow version containing that change (or an equivalent local patch) before merging.

I also found an MCP input-integrity issue: `duration` is declared as a JSON number but is read through `RequireInt`, whose mcp-go v0.56.0 implementation converts `float64` directly to `int`. A value such as `86400.5` therefore becomes `86400` and passes the duration whitelist instead of being rejected. The inline comment suggests reusing strict integer parsing.

No pull-request workflow run is available for this head SHA; the only combined status currently exposed is CodeRabbit.

kurotsuke:
On the duration comment: fixed in `e149120` — it now goes through the same strict int64 path as `start_time`/`end_time` with a bounds check before the int conversion.

On the whatsmeow blocker: agreed. The framing change is in tulir/whatsmeow#1214, which I opened for exactly this reason — it's still awaiting review upstream. I'll bump the whatsmeow pin here as soon as it lands. I've verified end-to-end delivery (rendering + RSVP on recipient devices) running this branch against a whatsmeow build that includes #1214. Happy to mark this PR as draft until the dependency merges if you prefer.

PR #797, opened 2026-08-16
docs: correct spelling in readme.md

vladdoster:
# PR Title

Correct spelling in `readme.md`

## Context

Fix error to match rest of examples

## Test Results

N/A

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -428,13 +428,13 @@ multi-device setups.
 Using Docker Hub:
 
 ```bash
-docker run --detach --publish=3000:3000 --name=whatsapp --restart=always --volume=$(docker volume create --name=whatsapp):/app/storages aldinokemal2104/go-whatsapp-web-multidevice rest --autoreply="Dont't reply this message please"
+docker run --detach --publish=3000:3000 --name=whatsapp --restart=always --volume=$(docker volume create --name=whatsapp):/app/storages aldinokemal2104/go-whatsapp-web-multidevice rest --autoreply="Don't reply this message please"
 ```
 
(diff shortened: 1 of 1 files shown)


aldinokemal (commented):
Reviewed the full PR diff and current discussion.

No blocker found. This is a documentation-only change limited to `readme.md`, replacing the malformed `Dont't` string with `Don't` in both Docker run examples. The corrected apostrophe is safe inside the existing double-quoted `--autoreply` value, so the examples remain valid shell syntax and there is no Go/whatsmeow/protobuf/runtime behavior change.

I also checked the existing CodeRabbit MD046 note. The fenced code-block style was already present before this PR and is used in the surrounding README; this PR does not introduce that formatting pattern, so I would not treat it as a blocker for this spelling fix.

CI note: there is no GitHub Actions PR workflow run for head `92e69d544bbb169de5774c82ca8ea7fe70e4d787`; the only commit status currently present is CodeRabbit success. For this README-only two-line text correction, I don't see a missing validation that should block merge.

aldinokemal (approved):
Re-reviewed the full current diff. This PR only fixes the `Dont't` typo to `Don't` in the two Docker `--autoreply` examples. The apostrophe remains safely inside a double-quoted shell argument, so the examples retain their intended command semantics.

The existing MD046 thread is about the surrounding README code-block style, not a regression introduced by these two changed lines. No runtime/API/protocol behavior is touched, GitHub reports the PR mergeable, and there are no workflow runs attached to this documentation-only head.

No blocking issue found.

PR #802, opened 2026-08-21
feat(mcp): add embedded OAuth 2.1 authorization

### LINKED ISSUE #799: Feature proposal: embedded OAuth 2.1 for the MCP endpoint (claude.ai custom connector support)
## Problem

The MCP endpoint (`/mcp`) inherits the REST server's basic auth. That works for MCP clients that can send custom headers (Claude Code, Cursor, n8n), but **claude.ai web/desktop custom connectors only support OAuth** — there is currently no way to connect GOWA to claude.ai directly.

## Proposal

Embed a minimal OAuth 2.1 authorization server in the existing REST server, following the MCP authorization spec:

- **Opt-in:** `MCP_OAUTH_ENABLED=true` + `MCP_OAUTH_ISSUER_URL=<public https URL>`. With the flag off (default), behavior is completely unchanged.
- **Coexistence:** `/mcp` then accepts `Authorization: Bearer <token>` **or** the existing basic auth — no breaking change for current MCP users.
- **Self-contained:** no new Go dependencies, no external IdP. Discovery via RFC 9728/8414 metadata, RFC 7591 dynamic client registration, authorization-code flow with mandatory PKCE (S256), refresh-token rotation with reuse detection (family revocation).
- **Login identity:** the authorize page checks credentials against the existing `APP_BASIC_AUTH` pairs — no new credential store.
- **Persistence:** clients/codes/tokens in a new `storages/oauth.db` (SQLite, via the existing driver abstraction); tokens stored only as SHA-256 hashes.

## Status

I have a working, tested implementation ready (unit + full-flow integration tests, README docs including a claude.ai troubleshooting section for Cloudflare-fronted deployments). Happy to open the PR if you're open to the feature — and to adjust the design to your preferences first.

aldinokemal:
## Summary

Adds an opt-in embedded OAuth authorization server for the existing streamable-HTTP MCP endpoint so remote MCP clients such as Claude custom connectors can authenticate without requiring a custom Basic Auth header.

Closes #799.

## Behavior

- OAuth stays disabled by default with `MCP_OAUTH_ENABLED=false`.
- When enabled, `/mcp` accepts either OAuth Bearer tokens or the existing Basic Auth credentials.
- Normal REST/UI routes keep their current Basic Auth behavior.
- OAuth login reuses `APP_BASIC_AUTH`; there is no second user/credential store.
- OAuth state is stored in a dedicated `storages/oauth.db` by default using the repo's existing CGO/pure-Go SQLite abstraction.
- No new Go dependency, JWT signing key, or external IdP is introduced.

## OAuth / MCP protocol

Implements:

- RFC 9728 protected-resource metadata using the MCP resource URL path-insertion form
- RFC 8414 authorization-server metadata
- authorization-code flow with mandatory PKCE `S256`
- RFC 8707 `resource` binding for authorization, token issuance, refresh, and MCP access-token validation
- RFC 9207 `iss` on successful authorization responses
- RFC 7591 Dynamic Client Registration for Claude/backward-compatible MCP clients
- exact registered redirect URI matching
- public clients only (`token_endpoint_auth_method=none`)

DCR is treated as a compatibility path and the server does **not** advertise Client ID Metadata Document support yet, matching the current MCP direction without breaking Claude compatibility.

## Token security

- cryptographically random opaque authorization codes/access tokens/refresh tokens
- only SHA-256 token/code hashes are persisted
- authorization codes are single-use and expire after 5 minutes
- access tokens expire after 1 hour
- refresh tokens rotate on every use
- reuse of a rotated refresh token revokes the entire token family, including access tokens
- authorization/login/token responses use `no-store`; the login page is frame-denied and uses a restrictive CSP
- the authorization page displays the registered redirect hostname before credentials are submitted

## Routing / backward compatibility

OAuth discovery, login, registration, token endpoints, and the OAuth-enabled MCP route are registered before the application's global Basic Auth middleware. This is necessary so OAuth clients can discover/authenticate and so Bearer tokens are not rejected by Fiber's Basic Auth middleware first.

When OAuth is disabled, MCP remains registered in its historical location behind the global Basic Auth middleware.

## Configuration

New settings:

```env
MCP_OAUTH_ENABLED=false
MCP_OAUTH_ISSUER_URL=
MCP_OAUTH_RESOURCE_URL=
MCP_OAUTH_DB_URI=file:storages/oauth.db
```

Equivalent CLI flags are included. `MCP_OAUTH_RESOURCE_URL` is optional and defaults to the issuer origin + `APP_BASE_PATH` + `/mcp`.

See `docs/mcp-oauth.md` for Claude custom connector setup, `APP_BASE_PATH` well-known routing, and Cloudflare/reverse-proxy notes.

## Tests added

Coverage was added for:

- RFC well-known metadata/path insertion
- DCR redirect validation
- complete authorization-code + PKCE flow
- Bearer authentication and Basic Auth fallback
- wrong resource and code replay
- refresh rotation + reuse/family revocation
- storage of token/code hashes rather than raw secrets
- access-token resource binding
- redirect-prefix attacks
- wrong PKCE verifier behavior
- native loopback redirect restrictions (loopback IP literals only; `localhost` rejected per RFC 8252 section 8.3)
- OAuth POST body cap coverage across every routing-path alias Fiber accepts (trailing slash, case, absolute-form) and non-interference with the app-wide `BodyLimit`

## Verification status

Verified on head `e33dba8` with Go 1.26.6:

```bash
cd src
go build ./...                                              # ok
go vet ./...                                                # ok
go test ./...                                               # all packages pass
golangci-lint v2.12.2 run --new-from-rev=origin/main ./...   # 0 issues
```

`gofmt -l` is clean for every file this PR touches. The four files it still lists
(`pkg/error/app_error.go`, `pkg/utils/whatsapp.go`,
`infrastructure/whatsapp/chatwoot_content_test.go`, `usecase/device_test.go`) are
pre-existing and untouched by this branch.


src/ui/mcp/oauth/server.go
```diff
@@ -0,0 +1,609 @@
...
+	}
+	client := Client{
+		ID:                      clientID,
+		Name:                    name,
+		RedirectURIs:            append([]string(nil), req.RedirectURIs...),
+		ApplicationType:         req.ApplicationType,
+		TokenEndpointAuthMethod: req.TokenEndpointAuthMethod,
+		CreatedAt:               s.now(),
+	}
+	if err := s.store.createClient(c.Context(), client); err != nil {
```
aldinokemal: 🔴 **Blocker — public DCR is currently an unbounded persistent-write DoS.** This route is registered before Basic Auth, so anyone can reach it. The app-level Fiber `BodyLimit` is tied to `WhatsappSettingMaxVideoSize` (100 MB), while each `redirect_uri` has no length bound and every successful request permanently inserts a new client row. An unauthenticated caller can therefore POST very large but valid HTTPS redirect strings repeatedly and fill `storages/oauth.db` / the shared storage volume. Please add an OAuth-specific small body cap, per-field/aggregate metadata length limits, and a bounded registration policy (rate limit/quota and/or expiry/GC). Add a regression test proving oversized metadata is rejected before it reaches durable storage.

src/ui/mcp/oauth/server.go
```diff
@@ -0,0 +1,609 @@
...
+
+func (s *Server) validateAuthorizationRequest(c fiber.Ctx, req *authorizationRequest) (Client, error) {
+	if req.ResponseType != "code" {
+		return Client{}, oauthError(c, fiber.StatusBadRequest, "unsupported_response_type", "response_type must be code")
+	}
+	client, err := s.store.getClient(c.Context(), req.ClientID)
+	if err != nil {
+		return Client{}, oauthError(c, fiber.StatusBadRequest, "invalid_request", "unknown client_id")
+	}
+	if !containsExact(client.RedirectURIs, req.RedirectURI) {
```
aldinokemal: 🔴 **Blocker — exact matching breaks native loopback redirects with ephemeral ports.** `validRedirectURI` explicitly supports native `http://127.0.0.1:<port>` / loopback redirects, but RFC 8252 §7.3/§8.4 (carried into OAuth 2.1) requires the authorization server to allow the port used in the authorization request to differ for loopback IP redirect URIs. `containsExact` rejects that required case. Match registered native loopback redirects exactly except for the port (scheme + loopback address + path/query still must match), then keep storing the *actual* request redirect URI in the authorization code so the token exchange can continue to require an exact match to that value. Please add a test that registers one loopback port and authorizes with another.

aldinokemal (commented):
Verdict: REQUEST CHANGES

I reviewed the full OAuth surface (routing/config, discovery metadata, DCR, authorization-code + PKCE, resource binding, token/refresh storage, Basic fallback, tests/docs) against current `main` and the current MCP/OAuth requirements. The overall layering is clean and the successful auth-code path is thoughtfully implemented, but I would not mark this draft ready yet.

Two blocking issues are called out inline:

1. The unauthenticated DCR endpoint is an unbounded durable-write surface. It inherits GOWA's ~100 MB global request body limit, individual redirect URIs have no length bound, registrations never expire, and there is no per-source/client quota or cleanup. A remote unauthenticated caller can grow `storages/oauth.db` until the shared storage volume is exhausted. The current MCP move toward CIMD is motivated in part by avoiding unbounded registration state, so the DCR compatibility path needs explicit abuse bounds.
2. Native loopback redirect matching is too strict. RFC 8252 / OAuth 2.1 require authorization servers to allow the runtime port to differ for loopback IP redirect URIs. Registration currently accepts native loopback redirects, but authorization then requires byte-for-byte URI equality, which can reject desktop/CLI clients that bind an ephemeral port after registration.

The head is mergeable but still draft, and there are no GitHub Actions runs for `83b6df484a58103ca0eaa8cae478796c64ef3afa`; the PR itself also says the Go 1.25.5 suite was not executed. After fixing the two blockers, please add regression tests for DCR size/abuse bounds and loopback-port matching, then run `gofmt`, `go test ./...`, and `go vet ./...` on the exact head before marking ready.

src/ui/mcp/oauth/server.go
```diff
@@ -0,0 +1,693 @@
...
+	}
+	resource := &url.URL{Scheme: issuer.Scheme, Host: issuer.Host}
+	resource.Path = path.Join("/", strings.TrimSpace(appBasePath), "mcp")
+	return resource.String(), nil
+}
+
+func (s *Server) RegisterPublic(router fiber.Router) {
+	router.Get(wellKnownPath("oauth-authorization-server", s.issuer.Path), s.authorizationServerMetadata)
+	router.Get(wellKnownPath("oauth-protected-resource", s.resource.Path), s.protectedResourceMetadata)
+	router.Post(s.issuerEndpointPath("/oauth/register"), s.registerClient)
```
aldinokemal: 🔴 **Blocker — the public OAuth POST routes still inherit GOWA's 100 MB app-wide body limit.** The 16 KiB `len(c.Body())` check in `registerClient` happens inside the handler, after Fiber/fasthttp has accepted the request under the app `BodyLimit`, and `/oauth/authorize` plus `/oauth/token` have no small OAuth-specific guard at all. Since these routes are intentionally registered before Basic Auth, an unauthenticated caller can repeatedly send very large bodies and consume request memory/bandwidth even though DCR later returns 413. Put the OAuth POST endpoints behind a small request-size guard/limit before binding/processing (and keep the field-level checks as defense in depth). Add coverage for register/authorize/token oversized bodies.

src/ui/mcp/oauth/server.go
```diff
@@ -0,0 +1,693 @@
...
+	if req.ApplicationType != "web" && req.ApplicationType != "native" {
+		return oauthError(c, fiber.StatusBadRequest, "invalid_client_metadata", "unsupported application_type")
+	}
+	if req.TokenEndpointAuthMethod == "" {
+		req.TokenEndpointAuthMethod = "none"
+	}
+	if req.TokenEndpointAuthMethod != "none" {
+		return oauthError(c, fiber.StatusBadRequest, "invalid_client_metadata", "only public clients are supported")
+	}
+	if !onlySupported(req.ResponseTypes, "code") || !onlySupported(req.GrantTypes, "authorization_code", "refresh_token") {
```
aldinokemal, line 312: 🟠 **Major — `grant_types` / `response_types` are accepted as registration metadata but are not actually registered or enforced.** `Client` and `oauth_clients` persist neither field, `validateAuthorizationRequest` always permits the code flow, code exchange always issues a refresh token, and `/oauth/token` always accepts `refresh_token`. So a client that registers `grant_types:["authorization_code"]` gets a response claiming no refresh grant but still receives/uses a refresh token; a client registering only `refresh_token` can still run the authorization-code flow. RFC 7591 defines `grant_types` as the grants the client may use and requires the registration response to return the metadata actually registered. Either normalize every registration to this server's fixed `authorization_code + refresh_token`/`code` capabilities and return those normalized values, or persist the requested/accepted metadata and enforce it at authorization/token issuance. Also reject inconsistent grant/response combinations and add regression tests.

aldinokemal (commented):
Verdict: REQUEST CHANGES

Re-reviewed current head `b5fda30` after the previous two owner blockers were resolved. The loopback ephemeral-port fix is now correct: authorization allows a port change only for native loopback IP redirects, while the actual redirect URI is stored on the code and token exchange still requires that exact URI. The DCR persistence issue is also substantially improved with bounded metadata, durable client quotas, expiry cleanup, and token/code cleanup.

I still found two merge-blocking issues in the current implementation, called out inline:

1. The public OAuth POST surface still inherits the app-wide 100 MB request-body limit. The new 16 KiB registration check runs after Fiber has already accepted the request body, and `/oauth/authorize` + `/oauth/token` have no OAuth-specific size guard at all. Because these routes sit before Basic Auth, this adds an unauthenticated large-body resource-exhaustion surface. Apply a small request-size limit/guard to the OAuth POST routes before binding/processing, not only a `len(c.Body())` check inside DCR.
2. DCR `grant_types` / `response_types` are validated and echoed but never stored or enforced. A client can register metadata that says it may use only `authorization_code` (or even only `refresh_token`), yet the server always allows the code flow, always issues a refresh token, and accepts the refresh grant. RFC 7591 defines these as registered capabilities; the registration response must reflect what the server actually registered and the endpoints must enforce it.

Current MCP `2026-07-28` still permits DCR for backwards compatibility (though it is deprecated in favor of CIMD), and the PR's `application_type`, RFC 9207 `iss`, RFC 9728 discovery, RFC 8707 resource binding, PKCE S256, token hashing/rotation/reuse revocation, and Basic fallback all look directionally correct.

There are also no GitHub Actions runs for this head. CodeRabbit currently reports concrete golangci-lint failures in the new tests (`bodyclose` in `cmd/rest_test.go` and `noctx` in `oauth/security_test.go`); those should be cleaned up and the exact Go 1.25.5 `go test ./...` / `go vet ./...` suite should pass before merge.

src/cmd/rest_test.go
```diff
@@ -83,6 +83,37 @@ func TestCORSPreflightAllowsUIHeaders(t *testing.T) {
...
+	require.NoError(t, err)
+	assert.Equal(t, fiber.StatusOK, resp.StatusCode)
+	assert.Empty(t, resp.Header.Get("Access-Control-Allow-Origin"))
+
+	preflight := httptest.NewRequest("OPTIONS", "/gowa/oauth/authorize", nil)
+	preflight.Header.Set("Origin", "https://client.example.com")
+	preflight.Header.Set("Access-Control-Request-Method", "GET")
+	resp, err = app.Test(preflight)
+	require.NoError(t, err)
+	assert.Empty(t, resp.Header.Get("Access-Control-Allow-Origin"))
```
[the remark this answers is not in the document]
aldinokemal: Fixed in fd47564: both app.Test response bodies are now closed and asserted, with separate response variables so the first body is not lost before reassignment.

src/ui/mcp/oauth/server.go
aldinokemal: Fixed in fd47564 by normalizing DCR to the actual fixed server capabilities: every successful registration returns authorization_code + refresh_token and response type code. Supported-value validation remains in place, and a regression test covers a subset request being normalized honestly.

aldinokemal:
Addressed the remaining review feedback in fd47564. In addition to the four inline threads, this includes context-aware SQL initialization, a shared client insert helper, asserted rows.Close cleanup, removal of duplicate OAuth env loading, defensive redirect parsing, and reverse-proxy guidance for per-IP authorization login throttling/logging. Validation: go test ./..., go vet ./..., and golangci-lint v2.12.2 --new-from-rev=origin/main all pass (0 new issues). The repository-wide linter still reports 59 pre-existing issues outside this PR.

aldinokemal:
@coderabbitai review

src/ui/mcp/oauth/server.go
```diff
@@ -0,0 +1,720 @@
...
+		var cfg fasthttp.RequestConfig
+		if previous != nil {
+			cfg = previous(header)
+		}
+		if !bytes.Equal(header.Method(), []byte(fiber.MethodPost)) {
+			return cfg
+		}
+		requestPath, _, _ := bytes.Cut(header.RequestURI(), []byte{'?'})
+		for _, oauthPath := range paths {
+			if bytes.Equal(requestPath, oauthPath) &&
```
aldinokemal: [P1] The early 16 KiB cap is still bypassable because this compares the **raw request URI path byte-for-byte**, while this app leaves Fiber at its defaults (`StrictRouting=false`, `CaseSensitive=false`). A request such as `POST /oauth/token/` (or `/OAUTH/TOKEN`) can therefore route to the same `/oauth/token` handler but miss this comparison and fall back to the application's global body limit (~100 MB). `/oauth/authorize` has the same issue; `/oauth/register` only checks `len(c.Body())` after the large body has already been buffered. This reopens the unauthenticated large-body resource-exhaustion surface the previous fix was meant to close. Please make the header-time matcher cover every path form Fiber accepts (or make these routes strict/case-sensitive at the routing boundary) and add oversized-body regressions for at least `/oauth/token/` and a case variant.

aldinokemal (commented):
Verdict: REQUEST CHANGES

Re-reviewed the current head after the previous OAuth fixes. The earlier blockers are addressed: DCR now has durable quotas/cleanup, native loopback IP redirects allow an ephemeral port while binding the actual redirect URI into the code, the public OAuth POSTs gained an early fasthttp body cap, and DCR now returns the server's fixed grant/response capabilities. PKCE S256, RFC 8707 resource binding, RFC 9207 `iss`, one-time code exchange, refresh rotation/family revocation, hash-only token persistence, discovery metadata, and Basic-Auth fallback all still look coherent with the current MCP authorization direction.

There is one remaining blocking bypass in the new body-cap enforcement, noted inline. Because current Fiber routing is case-insensitive and non-strict by default, raw-path byte equality does not cover every URL that reaches these handlers, so unauthenticated requests can still fall back to the application's ~100 MB global body limit.

Current `main` is exactly this PR's base (`0427b9f`), now on Go 1.26.0 and whatsmeow `33cfac511629`; this PR does not touch the WhatsApp protocol boundary. I also found no GitHub Actions run for head `fd475648`. After fixing the route-alias cap bypass, please add regression coverage for accepted trailing-slash/case variants and run `go test ./...` + `go vet ./...` on the exact head. The PR description's verification text is also stale (it still says draft/Go 1.25.5), but that is documentation cleanup rather than a blocker.

src/ui/mcp/oauth/server.go
aldinokemal: Fixed in e33dba8. Reproduced first: with `fiber.New()` defaults, `POST /oauth/token/`, `/OAUTH/TOKEN`, and `/oauth/Token/` all reached the token handler (400 from the handler, not 404) while escaping the 16 KiB cap and falling back to the app-wide limit; `/oauth/register/` reached the handler and only hit the in-handler 413 after the body was buffered.

The matcher no longer touches the raw URI. It now parses the target with `fasthttp.URI.Parse(header.Host(), header.RequestURI())` — the same call `Request.parseURI` makes, whose `PathOriginal()` is exactly what Fiber assigns to `c.pathOriginal` in `Reset` — and then applies Fiber's own `configDependentPaths` normalization: ASCII lowercase (CaseSensitive off) and strip all trailing slashes when len > 1 (StrictRouting off). Registered paths go through the same normalization, so an issuer path with uppercase characters matches too. Using the shared parser also covers absolute-form targets and fragments, not just the query string the old `bytes.Cut` handled. On a parse error the cap is skipped, because `URI.parse` returns before setting `pathOriginal`, so Fiber sees an empty path and 404s rather than routing to an OAuth handler.

I did not change `StrictRouting`/`CaseSensitive`: both are app-wide `fiber.Config` fields in Fiber v3 with no per-route override, so flipping them would change matching for every existing REST/UI route.

Regression coverage in `TestOAuthBodyCapCoversRoutingPathAliases`: `/oauth/token/`, `/oauth/token//`, `/OAUTH/TOKEN`, `/oauth/Token/`, `/oauth/token?resource=...`, `http://gowa.example.com/oauth/token/`, plus trailing-slash and uppercase variants of `/oauth/authorize` and `/oauth/register` — each asserted to fail with `fasthttp.ErrBodyTooLarge` at the boundary. `TestOAuthBodyCapLeavesOtherRoutesAlone` asserts a non-OAuth POST route still gets the app-wide `BodyLimit`, so the cap did not widen past the OAuth surface. Verified red on the pre-fix matcher and green after.

Verification on e33dba8 with Go 1.26.6: `go build ./...`, `go vet ./...`, `go test ./...` all pass, and `golangci-lint v2.12.2 run --new-from-rev=origin/main ./...` reports 0 issues. `gofmt -l` is clean for every file this PR touches; the four files it still lists (`pkg/error/app_error.go`, `pkg/utils/whatsapp.go`, `infrastructure/whatsapp/chatwoot_content_test.go`, `usecase/device_test.go`) are pre-existing and untouched by this branch.

aldinokemal:
Addressed the two remaining items in e33dba8.

**1. Route-alias bypass of the OAuth body cap (owner P1).** Fixed — details and the reproduction in [the inline thread](https://github.com/aldinokemal/go-whatsapp-web-multidevice/pull/802#discussion_r3835472437). Short version: the header-time matcher now derives the path from the same `fasthttp.URI` parser Fiber uses and applies Fiber's own `configDependentPaths` normalization (ASCII lowercase, trailing slashes stripped), so every URL form Fiber routes to these handlers is capped. `StrictRouting`/`CaseSensitive` were deliberately left alone: they are app-wide in Fiber v3 with no per-route override, so changing them would alter matching for every existing REST/UI route.

**2. `localhost` native redirects (@coderabbitai, outside-diff).** Fixed. `validRedirectURI` now requires a loopback IP literal for `http` native redirects; the `strings.EqualFold(hostname, "localhost")` branch is gone. Two things made this worth taking rather than deferring: RFC 8252 §8.3 recommends the IP literal precisely because a hosts-file or resolver change can send the authorization code off the loopback interface, and `loopbackRedirectMatches` already gated the ephemeral-port exception on `net.ParseIP(...).IsLoopback()`, so `localhost` registrations were getting exact-match-only behavior anyway — accepting them was inconsistent with the port-flexibility the native path exists to provide. Nothing in the tree depended on it. `TestDCRNativeRedirectAllowsOnlyLoopbackHTTP` now also asserts `http://[::1]:PORT/callback` is accepted and that both `http://localhost:...` and `http://LocalHost:...` are rejected; `docs/mcp-oauth.md` states the requirement.

**Verification on e33dba8**, Go 1.26.6:

```
go build ./...                                              # ok
go vet ./...                                                # ok
go test ./...                                               # all packages pass
golangci-lint v2.12.2 run --new-from-rev=origin/main ./...   # 0 issues
```

Both new regressions were confirmed red on the pre-fix code and green after. `gofmt -l` is clean for every file this PR touches; the four files it still lists (`pkg/error/app_error.go`, `pkg/utils/whatsapp.go`, `infrastructure/whatsapp/chatwoot_content_test.go`, `usecase/device_test.go`) are pre-existing and untouched by this branch.

The stale verification section in the PR description is updated to match — this branch is no longer draft and the suite has been run on the exact head.

docs/mcp-oauth.md
```diff
@@ -0,0 +1,133 @@
...
+
+- authorization codes, access tokens, and refresh tokens are generated with `crypto/rand`;
+- only SHA-256 hashes are persisted;
+- authorization codes are single-use and expire after 5 minutes;
+- access tokens expire after 1 hour;
+- refresh tokens rotate on each use;
+- reusing a rotated refresh token revokes the whole token family;
+- PKCE with `S256` is mandatory;
+- redirect URIs must exactly match the URI registered by the client, except that native loopback-IP clients may change only the port used for their local callback;
+- redirect URIs must use HTTPS, except `native` clients, which may use `http` only with a loopback IP literal such as `http://127.0.0.1:PORT/callback` or `http://[::1]:PORT/callback`. `http://localhost:PORT/...` is rejected per RFC 8252 section 8.3, because name resolution can point `localhost` at a non-loopback interface;
```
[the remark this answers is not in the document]
aldinokemal: Obsolete as of cf201ec, which landed after this review ran on e33dba8. The sentence this objected to no longer exists: `localhost` is now accepted rather than rejected, and the RFC 8252 attribution is gone. Line 110 currently reads:

> redirect URIs must use HTTPS, except `native` clients, which may use `http` with a loopback IP literal such as `http://127.0.0.1:PORT/callback` or `http://[::1]:PORT/callback`. For Claude Code compatibility, the exact hostname `localhost` is also accepted only with an explicit port; unlike loopback IP literals, a registered `localhost` redirect receives no dynamic-port exception and must match exactly during authorization;

`grep -rn "RFC 8252\|8252" docs/mcp-oauth.md src/ui/mcp/oauth/` returns nothing on the current head, so there is no remaining RFC-mandate claim to soften. The point behind the comment was fair — §8.3 says `localhost` is NOT RECOMMENDED, not prohibited — and that is precisely why the branch moved to accepting it under narrow constraints instead of rejecting it outright.

The behavior is enforced as documented: `validRedirectURI` accepts the exact hostname `localhost` only for `application_type=native` over `http` and only with an explicit port, and `loopbackRedirectMatches` still gates the dynamic-port exception on `net.ParseIP(...).IsLoopback()`, so a `localhost` registration must match byte-for-byte at authorization. `TestDCRNativeRedirectAllowsOnlyLoopbackHTTP` covers the accepted forms (`127.0.0.1`, `[::1]`, `localhost`, `LocalHost`) and the rejected ones (`example.com`, `localhost.example.com`, `localhost` with no port, and `localhost` under `application_type=web`).

Verified on cf201ec with Go 1.26.6: `go vet ./...` and `go test ./...` pass, `golangci-lint v2.12.2 run --new-from-rev=origin/main ./...` reports 0 issues.

src/ui/mcp/oauth/security_test.go
```diff
@@ -0,0 +1,361 @@
...
+	assert.Equal(t, fiber.StatusBadRequest, register(`{"client_name":"Lookalike","redirect_uris":["http://localhost.example.com:49152/callback"],"application_type":"native","token_endpoint_auth_method":"none"}`))
+	assert.Equal(t, fiber.StatusBadRequest, register(`{"client_name":"No port","redirect_uris":["http://localhost/callback"],"application_type":"native","token_endpoint_auth_method":"none"}`))
+	assert.Equal(t, fiber.StatusBadRequest, register(`{"client_name":"Web","redirect_uris":["http://localhost:49152/callback"],"application_type":"web","token_endpoint_auth_method":"none"}`))
+}
+
+func TestDCRAcceptsClaudeCodeNativeLocalhostRegistration(t *testing.T) {
+	_, app := newOAuthTestServer(t, testIssuer, testResource)
+	body := `{"client_name":"Claude Code (gowa)","redirect_uris":["http://localhost:60390/callback"],"grant_types":["authorization_code","refresh_token"],"response_types":["code"],"token_endpoint_auth_method":"none","application_type":"native","scope":"mcp"}`
+	req := httptest.NewRequest("POST", "/oauth/register", strings.NewReader(body))
```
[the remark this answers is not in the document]
aldinokemal: Fixed in a71fc75. Updated both request constructors introduced by the Claude DCR regression to use httptest.NewRequestWithContext with t.Context(), preserving the request methods, targets, bodies, and coverage. The OAuth native and pure-Go suites, the full Go suite, go vet, and diff checks pass.

src/ui/mcp/oauth/server.go
```diff
@@ -0,0 +1,777 @@
...
+		return false
+	}
+	hostname := u.Hostname()
+	// Claude Code registers its native callback as localhost with an explicit
+	// ephemeral port. Accept that narrow compatibility form, but keep the
+	// authorization-time port exception limited to IP literals in
+	// loopbackRedirectMatches so localhost redirects remain exact matches.
+	if strings.EqualFold(hostname, "localhost") {
+		return u.Port() != ""
+	}
```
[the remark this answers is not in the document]
aldinokemal: Not applying this suggestion. The hostname-resolution risk is real, but Claude Code 2.1.239 was captured registering application_type native with http://localhost:<ephemeral-port>/callback; removing this branch reproduces the HTTP 400 and breaks claude mcp login. The compatibility exception is limited to native HTTP callbacks whose hostname is exactly localhost and that include an explicit port. Localhost authorization redirects remain byte-exact, including the port and path; the ephemeral-port matching exception remains limited to loopback IP literals. PKCE S256 is mandatory, while non-loopback HTTP and localhost lookalikes remain rejected with regression coverage. Leaving this thread open for maintainer judgment on the compatibility tradeoff.

aldinokemal (commented):
Verdict: APPROVE

Re-reviewed the current head after the previous fixes. The blocking issues from the earlier rounds are resolved: native loopback redirects now handle ephemeral ports without weakening the redirect binding, DCR has bounded request/storage growth, the OAuth POST body limit is enforced at the fasthttp boundary and now covers Fiber's case-insensitive/non-strict route aliases, and the registration response reflects the server's actual fixed capabilities.

I also re-checked the security-critical paths: PKCE S256, one-time authorization-code exchange, exact resource binding, hash-only secret persistence, refresh-token rotation with family revocation on reuse, Basic Auth fallback isolation, and OAuth-only MCP auth remain consistent and transactional. I don't see a remaining merge blocker in this head.

There is no GitHub Actions run attached to this exact head, so the merge still relies on the PR's documented local build/vet/test/lint verification.

PR #808, opened 2026-08-22
Upgrade Go to 1.26 and update infrastructure configuration

bektigalan:
- Update Go version to 1.26 in workflows, Dockerfiles, and documentation
- Change Docker network mode to host
- Fix potential CRLF issues in entrypoint script

# PR Title

## Context

## Test Results

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,8 +9,7 @@ services:
       context: .
       dockerfile: ./docker/golang.Dockerfile
     restart: "on-failure"
-    ports:
-      - "3000:3000"
+    network_mode: "host"
     env_file:
       - src/.env
     volumes:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version: "1.25"
+          go-version: "1.26"
 
       - name: Install cross-compilation toolchains
         run: |
@@ -163,7 +163,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version: "1.25"
+          go-version: "1.26"
 
       - name: Install GoReleaser
         uses: goreleaser/goreleaser-action@v7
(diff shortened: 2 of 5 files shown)


docker-compose.yml
```diff
@@ -9,8 +9,7 @@ services:
       context: .
       dockerfile: ./docker/golang.Dockerfile
     restart: "on-failure"
-    ports:
-      - "3000:3000"
+    network_mode: "host"
```
aldinokemal, line 12: **Blocking portability/security regression:** please don't make `network_mode: "host"` the default Compose configuration. The app already binds `0.0.0.0:3000`, so the existing bridge network plus `3000:3000` publishes it correctly while preserving Docker network isolation. Host mode additionally requires an opt-in setting on Docker Desktop 4.34+ and makes the container contend directly for host port 3000. If host networking is needed for a specific environment, expose it through an optional override/profile rather than replacing the portable default.

aldinokemal (changes requested):
Reviewed the complete infrastructure-only change against current `main` (`src/go.mod` already requires Go 1.26.0), the release/Docker build paths, current Compose behavior, and available checks.

The Go 1.26 alignment is correct and needed: the module already declares Go 1.26.0, `golang:1.26-alpine3.23` is a valid official image tag, and moving the release jobs/build image from 1.25 to 1.26 removes an unnecessary toolchain mismatch. The entrypoint CRLF normalization is also safe as a build-time defensive step.

One blocking deployment regression remains: making host networking the default Compose mode changes a portable, isolated `3000:3000` service into a host-network service. On Docker Desktop, host networking is opt-in (4.34+), and on Linux it removes network namespace isolation and requires host port 3000 to be free. The application does not need host networking for outbound WhatsApp traffic. Keep the existing bridge + published-port default; if direct host access is needed for a special deployment, provide it as an opt-in Compose override/profile instead of changing the default.

There are no GitHub Actions workflow runs attached to head `c12a239f944874c8b567d2a4d4dcdff6facd8ef9`, so after fixing the Compose default please at least validate `docker compose config`, `docker compose build`, container startup/health, and the release build path under Go 1.26.

PR #809, opened 2026-08-22
feat(send): add HD photo and video quality

### LINKED ISSUE #807: ⚡️ Sending HD media
### What problem does your feature request solve?

I would like to request an HD media feature for sending photos and videos. I tried setting compress = false in the current version, but the media sent via WhatsApp is still low quality. For example, text inside image reports remains very difficult to read. Thanks.

### Describe the solution you are proposing.

The pictures below are downloaded from Whatsapp but original still HD.

Image

Image

### Describe alternatives you have considered.

_No response_

### Importance

Nice to Have

### Additional Context

- [x] This missing feature presents as a bug.
- [ ] This missing feature is forcing me to consider alternatives.

aldinokemal:
## Context

- Add an explicit `hd` option to the REST and MCP image/video send contracts for #807.
- Keep all existing defaults unchanged: legacy image/video compression still uses its current profile, while `compress=false` still sends the original bytes.
- Make `hd=true` take precedence over `compress` so clients do not have to coordinate conflicting quality switches.

## HD behavior

- Photos are resized only when necessary to fit a 2560px maximum edge and encoded at JPEG quality 92, matching current WhatsApp Web HD photo parameters without upscaling smaller images.
- Videos use an aspect-preserving H.264 CRF 23 profile capped inside 1280x1280, producing 1280x720 for normal landscape video and 720x1280 for portrait video without upscaling smaller sources.
- Image width/height and video width/height/duration are populated in the outgoing protobuf metadata.
- OpenAPI and the consolidated `whatsapp_send` MCP schema document the new option.

## Test Results

- Confirmed each regression test failed before its corresponding implementation (missing request field, quality selection, and metadata fallback).
- `go test ./... -count=1`
- `go vet ./...`
- `go build -o /tmp/issue807-whatsapp .`
- `gofmt -l` on every changed Go file (no output)
- `git diff --check origin/main...HEAD`
- Exercised the FFmpeg HD filter in the project's runtime image:
  - 1920x1080 -> 1280x720
  - 1080x1920 -> 720x1280
  - 640x360 -> 640x360 (no upscaling)

Closes #807

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -1310,6 +1310,10 @@ paths:
                   type: boolean
                   example: false
                   description: Compress image
+                hd:
+                  type: boolean
+                  example: true
+                  description: Send a high-definition image rendition (maximum edge 2560px). Overrides compress when true.
                 duration:
                   type: integer
                   example: 86400
@@ -1555,6 +1559,10 @@ paths:
                   type: boolean
                   example: false
                   description: Compress video
+                hd:
+                  type: boolean
+                  example: true
+                  description: Send a high-definition 720p-class video rendition. Overrides compress when true.
                 gif_playback:
                   type: boolean
                   example: false
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -1562,7 +1562,7 @@ paths:
                 hd:
                   type: boolean
                   example: true
-                  description: Send a high-definition 720p-class video rendition. Overrides compress when true.
+                  description: Send an H.264 high-definition video rendition at CRF 23, capped at 1280x1280 while preserving aspect ratio and avoiding upscaling. Overrides compress when true.
                 gif_playback:
                   type: boolean
                   example: false
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -1562,7 +1562,7 @@ paths:
                 hd:
                   type: boolean
                   example: true
-                  description: Send an H.264 high-definition video rendition at CRF 23, capped at 1280x1280 while preserving aspect ratio and avoiding upscaling. Overrides compress when true.
+                  description: Send an H.264 8-bit YUV 4:2:0 high-definition video rendition at CRF 23, capped at 1280x1280 while preserving aspect ratio and avoiding upscaling. Overrides compress when true.
                 gif_playback:
                   type: boolean
                   example: false
--- a/src/usecase/send.go
+++ b/src/usecase/send.go
@@ -62,6 +62,10 @@ func prepareImageForSend(src image.Image, compress, hd bool) (image.Image, bool)
 	return src, false
 }
 
+func openImageForSend(imagePath string, hd bool) (image.Image, error) {
+	return imaging.Open(imagePath, imaging.AutoOrientation(hd))
+}
+
 func saveProcessedImage(src image.Image, directory, imageName string, hd bool) (string, error) {
 	prefix := "new-"
 	var saveOptions []imaging.EncodeOption
@@ -81,6 +85,7 @@ func buildHDVideoFFmpegArgs(inputPath, outputPath string) []string {
 		"-crf", "23",
 		"-preset", "fast",
 		"-vf", "scale='min(1280,iw)':'min(1280,ih)':force_original_aspect_ratio=decrease:force_divisible_by=2",
+		"-pix_fmt", "yuv420p",
 		"-c:a", "aac",
 		"-b:a", "128k",
 		"-movflags", "+faststart",
@@ -392,7 +397,7 @@ func (service serviceSend) SendImage(ctx context.Context, request domainSend.Ima
 	deletedItems = append(deletedItems, oriImagePath)
 
 	/* Generate thumbnail with smalled image size */
-	srcImage, err := imaging.Open(oriImagePath)
+	srcImage, err := openImageForSend(oriImagePath, request.HD)
 	if err != nil {
 		return response, pkgError.InternalServerError(fmt.Sprintf("Failed to open image file '%s' for thumbnail generation: %v. Possible causes: file not found, unsupported format, or permission denied.", oriImagePath, err))
 	}
--- a/src/usecase/send.go
+++ b/src/usecase/send.go
@@ -62,6 +62,18 @@ func prepareImageForSend(src image.Image, compress, hd bool) (image.Image, bool)
 	return src, false
 }
 
+func saveProcessedImage(src image.Image, directory, imageName string, hd bool) (string, error) {
+	prefix := "new-"
+	var saveOptions []imaging.EncodeOption
+	if hd {
+		prefix = "hd-"
+		imageName = strings.TrimSuffix(imageName, filepath.Ext(imageName)) + ".jpg"
+		saveOptions = append(saveOptions, imaging.JPEGQuality(92))
+	}
+	processedImagePath := filepath.Join(directory, prefix+imageName)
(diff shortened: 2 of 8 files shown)


src/usecase/send.go
```diff
@@ -281,19 +393,20 @@ func (service serviceSend) SendImage(ctx context.Context, request domainSend.Ima
...
+		if request.HD {
+			prefix = "hd-"
+			saveOptions = append(saveOptions, imaging.JPEGQuality(92))
 		}
-		newImage := imaging.Resize(openImageBuffer, 600, 0, imaging.Lanczos)
-		newImagePath := fmt.Sprintf("%s/new-%s", config.PathSendItems, imageName)
-		if err = imaging.Save(newImage, newImagePath); err != nil {
-			return response, pkgError.InternalServerError(fmt.Sprintf("failed to save image %v", err))
+		processedImagePath := fmt.Sprintf("%s/%s%s", config.PathSendItems, prefix, imageName)
+		if err = imaging.Save(preparedImage, processedImagePath, saveOptions...); err != nil {
```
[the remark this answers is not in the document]
aldinokemal: Fixed in 75ab1b4. HD processed images now use a .jpg output path, so imaging selects the JPEG encoder and applies quality 92. TestSaveProcessedImageUsesJPEGForHDPNG verifies that PNG input emits image/jpeg bytes.

docs/openapi.yaml
```diff
@@ -1555,6 +1559,10 @@ paths:
                   type: boolean
                   example: false
                   description: Compress video
+                hd:
+                  type: boolean
+                  example: true
+                  description: Send a high-definition 720p-class video rendition. Overrides compress when true.
```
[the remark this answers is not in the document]
aldinokemal: Updated in 75ab1b4. REST and MCP descriptions now state H.264 CRF 23, a 1280x1280 bounding box, preserved aspect ratio, no upscaling, and hd precedence over compress.

src/usecase/send_test.go
```diff
@@ -16,6 +18,141 @@ import (
...
+		{name: "small image is not upscaled", width: 1200, height: 800, wantWidth: 1200, wantHeight: 800},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := resizeImageForHD(image.NewRGBA(image.Rect(0, 0, tt.width, tt.height)))
+			if got.Bounds().Dx() != tt.wantWidth || got.Bounds().Dy() != tt.wantHeight {
+				t.Fatalf("resizeImageForHD() = %dx%d, want %dx%d", got.Bounds().Dx(), got.Bounds().Dy(), tt.wantWidth, tt.wantHeight)
+			}
+		})
```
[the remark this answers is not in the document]
aldinokemal: Updated in 75ab1b4. All new use-case assertions now use testify/assert, including equality and error checks.

src/usecase/send.go
```diff
@@ -281,19 +405,14 @@ func (service serviceSend) SendImage(ctx context.Context, request domainSend.Ima
...
-		// Resize image
-		openImageBuffer, err := imaging.Open(oriImagePath)
-		if err != nil {
-			return response, pkgError.InternalServerError(fmt.Sprintf("Failed to open image file '%s' for compression: %v. Possible causes: file not found, unsupported format, or permission denied.", oriImagePath, err))
-		}
-		newImage := imaging.Resize(openImageBuffer, 600, 0, imaging.Lanczos)
-		newImagePath := fmt.Sprintf("%s/new-%s", config.PathSendItems, imageName)
-		if err = imaging.Save(newImage, newImagePath); err != nil {
-			return response, pkgError.InternalServerError(fmt.Sprintf("failed to save image %v", err))
+	preparedImage, processed := prepareImageForSend(srcImage, request.Compress, request.HD)
```
aldinokemal, line 413: [P1] HD re-encoding loses EXIF orientation and can send phone photos sideways. `srcImage` is opened above with plain `imaging.Open(oriImagePath)`, while imaging v1.6.2 disables AutoOrientation by default. When `hd=true`, this branch always re-encodes to a new JPEG, which drops the original EXIF Orientation tag without first rotating the pixels. A portrait JPEG that relies on Orientation=6/8 therefore becomes physically rotated, and the newly populated width/height describe that wrong orientation. Please decode the image used for re-encoding with `imaging.AutoOrientation(true)` (or otherwise apply EXIF orientation before saving), and add a regression fixture/test with a non-1 EXIF orientation.

src/usecase/send.go
```diff
@@ -38,6 +40,128 @@ import (
...
+		saveOptions = append(saveOptions, imaging.JPEGQuality(92))
+	}
+	processedImagePath := filepath.Join(directory, prefix+imageName)
+	return processedImagePath, imaging.Save(src, processedImagePath, saveOptions...)
+}
+
+func buildHDVideoFFmpegArgs(inputPath, outputPath string) []string {
+	return []string{
+		"-i", inputPath,
+		"-c:v", "libx264",
```
aldinokemal, line 84: [P1] Please force an interoperable 8-bit 4:2:0 pixel format for the new HD video path. `libx264` accepts YUV 4:2:2, 4:4:4 and 10-bit formats too; with no `-pix_fmt yuv420p`, the encoded H.264 profile/pixel format can depend on the source and FFmpeg build. That means a 10-bit/4:4:4 source can produce High10/High444 H.264 even though this path is supposed to create a WhatsApp-compatible HD rendition. Add `-pix_fmt yuv420p` (and an argument regression test; profile/level can also be made explicit if desired) so the output is deterministic across accepted input formats.

aldinokemal (commented):
Verdict: REQUEST CHANGES

The REST/MCP contract wiring, HD-vs-compress precedence, actual-byte dimension metadata, and pinned whatsmeow protobuf fields all look coherent. I found two media-processing blockers before I would consider this merge-safe, both on common real-world inputs. There is also no GitHub Actions run attached to this exact head, so please rerun the documented build/vet/test checks after fixing them.

The pinned whatsmeow version on current main (`33cfac511629`) does expose `VideoMessage.seconds/height/width`; I found no separate HD-quality wire flag in that pinned protobuf, so doing HD as a prepared-media rendition is consistent with the current protocol boundary.

src/usecase/send.go
aldinokemal: Fixed in 6e075b7. The HD path now decodes with imaging.AutoOrientation(true) before thumbnail generation, resizing, JPEG encoding, and dimension extraction. TestOpenImageForSendAppliesEXIFOrientationForHD builds an Orientation=6 JPEG fixture and verifies raw 2x3 pixels become displayed 3x2 pixels.

src/usecase/send.go
aldinokemal: Fixed in 6e075b7. The HD FFmpeg profile now forces -pix_fmt yuv420p and the literal argument regression test covers it. Runtime verification converted a 1920x1080 yuv444p source to High-profile 1280x720 yuv420p output.

aldinokemal:
Verification on head `6e075b7` after resolving the EXIF-orientation and video-pixel-format findings:

- `go test ./... -count=1` — pass
- `go vet ./...` — pass
- `go build -o /tmp/issue807-whatsapp .` — pass
- `gofmt -l` on changed Go files — no output
- `git diff --check` — pass
- Runtime FFmpeg check: 1920x1080 `yuv444p` input -> High-profile 1280x720 `yuv420p` output

All review threads are resolved. CodeRabbit, Kilo, and GitGuardian pass on this head.

PR #810, opened 2026-08-22
fix(message): sync deletes to chat storage

### LINKED ISSUE #806: revoke and delete report success but leave the message in chatstorage.db
**Environment:** GOWA v9.1.0, macOS (darwin/arm64), single paired device.

Revoking or deleting a message succeeds on WhatsApp's side but never updates GOWA's own `chatstorage.db`. The row stays, with full content, and `updated_at` is not even touched.

**Steps to reproduce:**
1. Send a text message and note the returned message ID.
2. Revoke it (delete for everyone). The API returns `Revoke success <jid>@s.whatsapp.net (server timestamp: ...)`.
3. Also try delete (delete for me). The API returns `Message <id> deleted`.
4. Re-read the chat via the messages endpoint, or query `chatstorage.db` directly.

**Expected:** After a successful revoke — or a delete-for-me — the stored row is removed, or at least flagged, so that the Chats UI and the messages API stop returning its content.

**Actual:** Both operations report success and leave the row completely untouched.

```
14:43:02Z  send    -> row created, updated_at = 14:43:02Z
14:44:37Z  revoke  -> API: "Revoke success ...@s.whatsapp.net (server timestamp: 2026-08-21 14:44:37 +0000 UTC)"
15:07:xxZ  delete  -> API: "Message <id> deleted"
           read    -> same row, full content, updated_at STILL 14:43:02Z
```

The unchanged `updated_at` is the tell: the revoke/delete handlers never write to the store at all.

The revoke itself works — the recipient's client shows "This message was deleted". Only GOWA's local copy is out of sync.

**Why this matters:** the Chats UI presents `chatstorage.db` as the conversation. A message that no longer exists for either party is still displayed in full, and there is no API path to get rid of it — the only remedy is a manual `DELETE FROM messages WHERE id = '<id>'` against the database while the server is running.

This is how I hit it: I sent a message to the wrong contact, revoked it within ~90 seconds, and the text was still sitting in the Chats view afterwards.

**Scope of this report:** I verified self-initiated revoke and delete only. I did not test whether an incoming `protocolMessage` REVOKE from the peer is applied to the store — if it shares the same gap, GOWA would be retaining message content that the other party deliberately withdrew, which seems worth checking as part of the same fix.

aldinokemal:
## Context

- Delete the active device's stored message after WhatsApp confirms revoke or delete-for-me.
- Apply incoming REVOKE protocol events to device-scoped chat storage.
- Remove reactions and edit history while retaining Chatwoot correlation for asynchronous delete sync.
- Return an explicit partial-success error if WhatsApp succeeds but local cleanup fails.

Fixes #806

## Test Results

- go test ./... -count=1
- go vet ./...
- go build -o /tmp/gowa-issue806-whatsapp .

### DIFF
--- a/src/infrastructure/chatstorage/sqlite_repository.go
+++ b/src/infrastructure/chatstorage/sqlite_repository.go
@@ -694,24 +694,26 @@ func (r *SQLiteRepository) loadMessageReactions(deviceID, chatJID string, messag
 	return nil
 }
 
-// DeleteMessage deletes a specific message
+// DeleteMessage deletes a specific message. Chatwoot links are intentionally
+// preserved so asynchronous revoke/delete forwarding can still resolve them.
 func (r *SQLiteRepository) DeleteMessage(id, chatJID string) error {
 	if _, err := r.db.Exec("DELETE FROM message_reactions WHERE message_id = ? AND chat_jid = ?", id, chatJID); err != nil {
 		return err
 	}
-	if _, err := r.db.Exec("DELETE FROM chatwoot_message_links WHERE wa_message_id = ? AND wa_chat_jid = ?", id, chatJID); err != nil {
+	if _, err := r.db.Exec("DELETE FROM message_edits WHERE original_message_id = ? AND chat_jid = ?", id, chatJID); err != nil {
 		return err
 	}
 	_, err := r.db.Exec("DELETE FROM messages WHERE id = ? AND chat_jid = ?", id, chatJID)
 	return err
 }
 
-// DeleteMessageByDevice deletes a specific message for a specific device
+// DeleteMessageByDevice deletes a specific message for a specific device.
+// Chatwoot links are intentionally preserved for asynchronous delete sync.
 func (r *SQLiteRepository) DeleteMessageByDevice(deviceID, id, chatJID string) error {
 	if _, err := r.db.Exec("DELETE FROM message_reactions WHERE message_id = ? AND chat_jid = ? AND device_id = ?", id, chatJID, deviceID); err != nil {
 		return err
 	}
-	if _, err := r.db.Exec("DELETE FROM chatwoot_message_links WHERE wa_message_id = ? AND wa_chat_jid = ? AND device_id = ?", id, chatJID, deviceID); err != nil {
+	if _, err := r.db.Exec("DELETE FROM message_edits WHERE original_message_id = ? AND chat_jid = ? AND device_id = ?", id, chatJID, deviceID); err != nil {
 		return err
 	}
 	_, err := r.db.Exec("DELETE FROM messages WHERE id = ? AND chat_jid = ? AND device_id = ?", id, chatJID, deviceID)
@@ -1686,6 +1688,10 @@ func (r *SQLiteRepository) CreateMessage(ctx context.Context, evt *events.Messag
 	// Store the full sender JID (user@server) to ensure consistency between received and sent messages
 	sender := normalizedSender.ToNonAD().String()
 
+	if revokedMessageID := extractRevokedMessageID(evt.Message); revokedMessageID != "" {
+		return r.DeleteMessageByDevice(deviceID, revokedMessageID, chatJID)
+	}
+
 	// Get appropriate chat name using pushname if available (device-scoped)
 	chatName := r.GetChatNameWithPushNameByDevice(deviceID, normalizedChatJID, chatJID, normalizedSender.User, evt.Info.PushName)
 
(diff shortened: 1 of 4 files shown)


src/infrastructure/chatstorage/sqlite_repository.go
```diff
@@ -694,24 +694,26 @@ func (r *SQLiteRepository) loadMessageReactions(deviceID, chatJID string, messag
...
+// Chatwoot links are intentionally preserved for asynchronous delete sync.
 func (r *SQLiteRepository) DeleteMessageByDevice(deviceID, id, chatJID string) error {
 	if _, err := r.db.Exec("DELETE FROM message_reactions WHERE message_id = ? AND chat_jid = ? AND device_id = ?", id, chatJID, deviceID); err != nil {
 		return err
 	}
-	if _, err := r.db.Exec("DELETE FROM chatwoot_message_links WHERE wa_message_id = ? AND wa_chat_jid = ? AND device_id = ?", id, chatJID, deviceID); err != nil {
+	if _, err := r.db.Exec("DELETE FROM message_edits WHERE original_message_id = ? AND chat_jid = ? AND device_id = ?", id, chatJID, deviceID); err != nil {
 		return err
 	}
 	_, err := r.db.Exec("DELETE FROM messages WHERE id = ? AND chat_jid = ? AND device_id = ?", id, chatJID, deviceID)
```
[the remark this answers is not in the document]
aldinokemal: Fixed in b87c4a8. Both DeleteMessage methods now execute reaction, edit-history, and message deletion through one SQLite transaction, with rollback on any failure and commit only after all three statements succeed. Added a trigger-based regression test that forces the final message delete to abort and verifies the message, reaction, and edit rows all remain. Full go test, go vet, and go build verification passes.

aldinokemal, line 719: **P1 — make the dependent deletes atomic.** These three statements are separate autocommit operations. If reactions are deleted successfully and the edit/message delete then fails, callers receive the intended partial-success error but the database has already lost part of the local record. `DeleteChat` / `DeleteChatByDevice` in this same repository already use `Begin` + rollback + commit for exactly this kind of dependent-row cleanup. Please do the same here (and in `DeleteMessage`), preserving the current device/chat predicates and intentionally not touching `chatwoot_message_links`. Add a rollback regression test by forcing a later statement to fail and asserting message + reactions + edits all remain.

aldinokemal (commented):
Verdict: REQUEST CHANGES

I reviewed the full current head (`eb36addcd597e3174d853c586114785cad2f32da`) against base/current-main behavior and the exact whatsmeow revision pinned by `src/go.mod` (`33cfac511629`). The protocol ordering is sound: whatsmeow's `BuildRevoke` constructs a `ProtocolMessage_REVOKE`, and `SendMessage` waits for the server acknowledgement before returning, so performing local cleanup only after the WhatsApp action succeeds is the right boundary. The incoming REVOKE path also correctly recognizes the unwrapped protocol message, and the new usecase tests cover remote-success vs remote-failure plus device isolation.

One blocking storage-integrity issue remains: message cleanup now spans `message_reactions`, `message_edits`, and `messages`, but both deletion methods execute those statements as independent autocommit operations. If the second/third statement fails (lock, I/O error, constraint/schema problem), the method returns an error after already committing earlier deletions. That leaves a half-deleted local record even though this PR explicitly treats local-cleanup failure as a partial-success condition. The repository's existing `DeleteChat`/`DeleteChatByDevice` methods already use a transaction for the same dependent-row cleanup pattern, so message deletion should have the same atomicity.

Please wrap each message cleanup in one transaction, keep the Chatwoot link intentionally untouched, and add a failure/rollback regression test that proves an injected failure leaves the message, reactions, and edit history consistently unchanged. After that I don't see another merge blocker in this change.

There are no GitHub Actions workflow runs attached to this head; the PR reports local `go test ./... -count=1`, `go vet ./...`, and build verification.

src/infrastructure/chatstorage/sqlite_repository.go
aldinokemal: Already addressed in current head b87c4a8; this comment reviewed the previous head eb36add. Both deletion methods now use one transaction with deferred rollback and commit after all three deletes, Chatwoot links remain untouched, and the new trigger-based regression test verifies a forced final-statement failure preserves message, reaction, and edit rows. Full go test, go vet, and go build verification passes; CodeRabbit has re-reviewed and approved b87c4a8.

PR #816, opened 2026-08-26
feat(webhook): expose poll vote selections

### LINKED ISSUE #815: ⚡️ Receive Poll selections via webhook
### What problem does your feature request solve?

Hey, GoWa is awesome!
I've connected it to a webhook. When someone reacts to a poll, a message is received but it contains no information regarding the option selected in the poll. Is it possible to retrieve the option? This would make it super easy to make interactive WhatsApp experiences.

This is what I'm getting upon interaction with a poll:

```json
{
  "device_id": "<number1>@s.whatsapp.net",
  "event": "message",
  "payload": {
    "chat_id": "<number2>@s.whatsapp.net",
    "chat_lid": "<chat_lid>@lid",
    "from": "<number2>@s.whatsapp.net",
    "from_lid": "<lid>@lid",
    "from_name": "‎<Me>",
    "id": "3B9D73BE37F8C5BEFE35",
    "is_from_me": false,
    "timestamp": "2026-08-26T13:09:57Z"
  },
  "session_id": "My WhatsApp Account"
}
```

### Describe the solution you are proposing.

Return the selected option from the poll as well as the full list of options.

### Describe alternatives you have considered.

_No response_

### Importance

I Need It

### Additional Context

- [ ] This missing feature presents as a bug.
- [x] This missing feature is forcing me to consider alternatives.

aldinokemal:
# feat(webhook): expose poll vote selections

## Context

- Decrypt poll vote updates with the pinned whatsmeow client.
- Persist device- and chat-scoped poll definitions from live messages, sent polls, and history sync.
- Return selected option names, lowercase hashes, full option lists, and explicit resolution states while preserving `event: "message"` compatibility.
- Degrade safely when definitions or original message secrets are unavailable.
- Handle current poll creation variants plus poll edit/add-option updates.

Closes #815

## Test Results

- `go test ./infrastructure/chatstorage ./infrastructure/whatsapp ./usecase ./pkg/utils`
- `go test -race ./infrastructure/chatstorage ./infrastructure/whatsapp ./usecase`
- `go test -tags purego ./infrastructure/chatstorage ./infrastructure/whatsapp ./usecase`
- `go test ./...`
- `go vet ./...`
- `go build -o /tmp/gowa-poll-webhook .`
- `go build -tags purego -o /tmp/gowa-poll-webhook-purego .`
- `git diff --check origin/main...HEAD`

All commands passed. The crypto integration tests exercise real whatsmeow message-secret storage and both PN and LID group vote decryption. A live WhatsApp smoke test was not run because no external test recipient was authorized.

### DIFF
--- a/src/infrastructure/chatstorage/sqlite_repository.go
+++ b/src/infrastructure/chatstorage/sqlite_repository.go
@@ -763,32 +763,54 @@ func (r *SQLiteRepository) UpsertPollDefinition(definition *domainChatStorage.Po
 		return fmt.Errorf("failed to marshal poll options: %w", err)
 	}
 	now := time.Now()
-	if definition.CreatedAt.IsZero() {
-		definition.CreatedAt = now
+	if definition.UpdatedAt.IsZero() {
+		definition.UpdatedAt = now
 	}
-	definition.UpdatedAt = now
+	if definition.CreatedAt.IsZero() {
+		definition.CreatedAt = definition.UpdatedAt
+	}
 
-	result, err := r.db.Exec(`
-		UPDATE poll_definitions
-		SET question = ?, options_json = ?, selectable_option_count = ?, version = ?, updated_at = ?
-		WHERE device_id = ? AND chat_jid = ? AND poll_message_id = ?
-	`, definition.Question, string(optionsJSON), definition.SelectableOptionCount, definition.Version, definition.UpdatedAt,
-		definition.DeviceID, definition.ChatJID, definition.PollMessageID)
+	tx, err := r.db.Begin()
 	if err != nil {
 		return err
 	}
-	rowsAffected, _ := result.RowsAffected()
-	if rowsAffected != 0 {
-		return nil
+	defer tx.Rollback()
+
+	var existingUpdatedAt time.Time
+	err = tx.QueryRow(`
+		SELECT updated_at FROM poll_definitions
+		WHERE device_id = ? AND chat_jid = ? AND poll_message_id = ?
+	`, definition.DeviceID, definition.ChatJID, definition.PollMessageID).Scan(&existingUpdatedAt)
+	switch {
+	case err == nil:
+		if !definition.UpdatedAt.After(existingUpdatedAt) {
+			return tx.Commit()
+		}
+		_, err = tx.Exec(`
+			UPDATE poll_definitions
+			SET question = ?, options_json = ?, selectable_option_count = ?, version = ?, updated_at = ?
+			WHERE device_id = ? AND chat_jid = ? AND poll_message_id = ?
+		`, definition.Question, string(optionsJSON), definition.SelectableOptionCount, definition.Version, definition.UpdatedAt,
+			definition.DeviceID, definition.ChatJID, definition.PollMessageID)
+		if err != nil {
+			return err
+		}
+		return tx.Commit()
+	case err != sql.ErrNoRows:
+		return err
 	}
-	_, err = r.db.Exec(`
+
+	_, err = tx.Exec(`
 		INSERT INTO poll_definitions (
 			device_id, chat_jid, poll_message_id, question, options_json,
 			selectable_option_count, version, created_at, updated_at
 		) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)
 	`, definition.DeviceID, definition.ChatJID, definition.PollMessageID, definition.Question, string(optionsJSON),
 		definition.SelectableOptionCount, definition.Version, definition.CreatedAt, definition.UpdatedAt)
-	return err
+	if err != nil {
+		return err
+	}
+	return tx.Commit()
 }
 
 // GetPollDefinition retrieves one poll definition using its full device/chat identity.
--- a/src/infrastructure/chatstorage/sqlite_repository.go
+++ b/src/infrastructure/chatstorage/sqlite_repository.go
@@ -253,6 +253,9 @@ func (r *SQLiteRepository) DeleteChat(jid string) error {
 	if _, err := tx.Exec("DELETE FROM chatwoot_message_links WHERE wa_chat_jid = ?", jid); err != nil {
 		return err
 	}
+	if _, err := tx.Exec("DELETE FROM poll_definitions WHERE chat_jid = ?", jid); err != nil {
+		return err
+	}
 
 	// Delete messages after dependent rows to keep cleanup explicit.
 	_, err = tx.Exec("DELETE FROM messages WHERE chat_jid = ?", jid)
@@ -284,6 +287,9 @@ func (r *SQLiteRepository) DeleteChatByDevice(deviceID, jid string) error {
 		return err
 	}
 	if _, err := tx.Exec("DELETE FROM chatwoot_message_links WHERE wa_chat_jid = ? AND device_id = ?", jid, deviceID); err != nil {
+		return err
+	}
+	if _, err := tx.Exec("DELETE FROM poll_definitions WHERE chat_jid = ? AND device_id = ?", jid, deviceID); err != nil {
 		return err
 	}
 
@@ -709,6 +715,9 @@ func (r *SQLiteRepository) DeleteMessage(id, chatJID string) error {
 	if _, err := tx.Exec("DELETE FROM message_edits WHERE original_message_id = ? AND chat_jid = ?", id, chatJID); err != nil {
 		return err
 	}
+	if _, err := tx.Exec("DELETE FROM poll_definitions WHERE poll_message_id = ? AND chat_jid = ?", id, chatJID); err != nil {
+		return err
+	}
 	if _, err := tx.Exec("DELETE FROM messages WHERE id = ? AND chat_jid = ?", id, chatJID); err != nil {
 		return err
 	}
@@ -730,7 +739,121 @@ func (r *SQLiteRepository) DeleteMessageByDevice(deviceID, id, chatJID string) e
 	if _, err := tx.Exec("DELETE FROM message_edits WHERE original_message_id = ? AND chat_jid = ? AND device_id = ?", id, chatJID, deviceID); err != nil {
 		return err
 	}
+	if _, err := tx.Exec("DELETE FROM poll_definitions WHERE poll_message_id = ? AND chat_jid = ? AND device_id = ?", id, chatJID, deviceID); err != nil {
+		return err
+	}
 	if _, err := tx.Exec("DELETE FROM messages WHERE id = ? AND chat_jid = ? AND device_id = ?", id, chatJID, deviceID); err != nil {
+		return err
+	}
+	return tx.Commit()
+}
+
+// UpsertPollDefinition stores the ordered poll option catalogue used to map
+// decrypted vote hashes back to human-readable names.
+func (r *SQLiteRepository) UpsertPollDefinition(definition *domainChatStorage.PollDefinition) error {
+	if definition == nil {
+		return fmt.Errorf("poll definition is required")
+	}
+	if strings.TrimSpace(definition.DeviceID) == "" || strings.TrimSpace(definition.ChatJID) == "" || strings.TrimSpace(definition.PollMessageID) == "" {
+		return fmt.Errorf("poll definition requires device_id, chat_jid, and poll_message_id")
+	}
+
+	optionsJSON, err := json.Marshal(definition.Options)
+	if err != nil {
+		return fmt.Errorf("failed to marshal poll options: %w", err)
+	}
+	now := time.Now()
+	if definition.CreatedAt.IsZero() {
+		definition.CreatedAt = now
+	}
+	definition.UpdatedAt = now
+
+	result, err := r.db.Exec(`
+		UPDATE poll_definitions
+		SET question = ?, options_json = ?, selectable_option_count = ?, version = ?, updated_at = ?
+		WHERE device_id = ? AND chat_jid = ? AND poll_message_id = ?
+	`, definition.Question, string(optionsJSON), definition.SelectableOptionCount, definition.Version, definition.UpdatedAt,
+		definition.DeviceID, definition.ChatJID, definition.PollMessageID)
+	if err != nil {
+		return err
+	}
+	rowsAffected, _ := result.RowsAffected()
+	if rowsAffected != 0 {
+		return nil
+	}
+	_, err = r.db.Exec(`
+		INSERT INTO poll_definitions (
+			device_id, chat_jid, poll_message_id, question, options_json,
+			selectable_option_count, version, created_at, updated_at
+		) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)
+	`, definition.DeviceID, definition.ChatJID, definition.PollMessageID, definition.Question, string(optionsJSON),
+		definition.SelectableOptionCount, definition.Version, definition.CreatedAt, definition.UpdatedAt)
+	return err
+}
+
+// GetPollDefinition retrieves one poll definition using its full device/chat identity.
+func (r *SQLiteRepository) GetPollDefinition(deviceID, chatJID, pollMessageID string) (*domainChatStorage.PollDefinition, error) {
+	var definition domainChatStorage.PollDefinition
+	var optionsJSON string
+	err := r.db.QueryRow(`
+		SELECT device_id, chat_jid, poll_message_id, question, options_json,
+			selectable_option_count, version, created_at, updated_at
+		FROM poll_definitions
+		WHERE device_id = ? AND chat_jid = ? AND poll_message_id = ?
+	`, deviceID, chatJID, pollMessageID).Scan(
+		&definition.DeviceID, &definition.ChatJID, &definition.PollMessageID, &definition.Question, &optionsJSON,
+		&definition.SelectableOptionCount, &definition.Version, &definition.CreatedAt, &definition.UpdatedAt,
+	)
+	if err == sql.ErrNoRows {
+		return nil, nil
+	}
+	if err != nil {
+		return nil, err
+	}
+	if err := json.Unmarshal([]byte(optionsJSON), &definition.Options); err != nil {
+		return nil, fmt.Errorf("failed to decode poll options for %s: %w", pollMessageID, err)
+	}
+	return &definition, nil
+}
+
+// AppendPollOption atomically appends a new option while treating a repeated
+// hash as an idempotent delivery of the same add-option event.
+func (r *SQLiteRepository) AppendPollOption(deviceID, chatJID, pollMessageID string, option domainChatStorage.PollOption) error {
+	tx, err := r.db.Begin()
+	if err != nil {
+		return err
+	}
+	defer tx.Rollback()
+
+	var optionsJSON string
+	err = tx.QueryRow(`
+		SELECT options_json FROM poll_definitions
+		WHERE device_id = ? AND chat_jid = ? AND poll_message_id = ?
+	`, deviceID, chatJID, pollMessageID).Scan(&optionsJSON)
+	if err == sql.ErrNoRows {
+		return fmt.Errorf("poll definition %s not found", pollMessageID)
+	}
+	if err != nil {
+		return err
+	}
+	var options []domainChatStorage.PollOption
+	if err := json.Unmarshal([]byte(optionsJSON), &options); err != nil {
+		return fmt.Errorf("failed to decode poll options for %s: %w", pollMessageID, err)
+	}
+	for _, existing := range options {
+		if existing.Hash == option.Hash {
+			return tx.Commit()
+		}
+	}
+	options = append(options, option)
+	encoded, err := json.Marshal(options)
+	if err != nil {
+		return fmt.Errorf("failed to marshal poll options: %w", err)
+	}
+	if _, err := tx.Exec(`
+		UPDATE poll_definitions SET options_json = ?, updated_at = ?
+		WHERE device_id = ? AND chat_jid = ? AND poll_message_id = ?
+	`, string(encoded), time.Now(), deviceID, chatJID, pollMessageID); err != nil {
 		return err
 	}
 	return tx.Commit()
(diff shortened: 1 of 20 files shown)


aldinokemal (commented):
Verdict: REQUEST CHANGES

The poll model/storage work is generally aligned with current main and the pinned whatsmeow revision, but there is one blocking lifecycle bug in the live-event path: poll decryption currently inherits the context captured when the whatsmeow event handler was registered. For clients created by the REST login/pairing paths, that is an HTTP request context and is canceled after the request completes. Poll votes/secret poll updates then attempt message-secret DB reads with an already-canceled context and degrade valid events to `decrypt_failed`.

Please detach poll processing from request cancellation while preserving device-scoped context values, and add a regression test that processes a real encrypted poll vote after canceling the original handler/request context.

I also checked the protocol-facing APIs against `go.mau.fi/whatsmeow` pinned at `33cfac511629`: `HashPollOptions`, `DecryptPollVote`, `DecryptSecretEncryptedMessage`, and historical message-secret storage all exist at that revision and the general poll hash/decryption approach is sound. Migration 44 is append-only against current main. No GitHub Actions workflow run is currently attached to this exact head, so please rerun the normal test/vet/build checks after the fix.

src/infrastructure/whatsapp/event_message_handler.go
```diff
@@ -32,13 +32,14 @@ func handleMessage(ctx context.Context, evt *events.Message, chatStorageRepo dom
 	// edit-handling paths unchanged. No-op when the envelope is absent or when
 	// decryption fails.
 	evt = materializeSecretEditMessage(ctx, evt, client)
+	pollPayload := preparePollWebhookPayload(ctx, client, chatStorageRepo, evt)
```
aldinokemal, line 35: [P1] Detach poll crypto/storage from the handler's captured request context. `EnsureClient(ctx, ...)` registers `client.AddEventHandler(func(rawEvt any) { handler(ctx, inst, rawEvt) })`, while `Login` / `LoginWithCode` pass their HTTP request context into `EnsureClient`. After that request returns, this `ctx` is canceled, but future WhatsApp events keep reusing it. This new call reaches `NormalizeJIDFromLID` and, for votes/secret poll updates, the pinned whatsmeow `DecryptPollVote` / `DecryptSecretEncryptedMessage`; at `33cfac511629` those call `Store.MsgSecrets.GetMessageSecret(ctx, ...)`. A valid vote on a client paired through REST can therefore be reported as `resolution_status=decrypt_failed` simply because the old login request ended (and LID normalization may also fail). Please run poll preparation under a detached, bounded context such as `context.WithTimeout(context.WithoutCancel(ctx), ...)` so DeviceFromContext values are preserved without request cancellation, or register the event handler with a device-lifecycle context. Add a regression test that cancels the original parent context before decrypting a real stored-secret vote.

aldinokemal: Fixed in 0e8edf0157838c6d2787982da8687547ee42abff. `preparePollWebhookPayload` now runs under `context.WithTimeout(context.WithoutCancel(ctx), 30*time.Second)` — the same pattern `handleWebhookForward` already uses — so LID normalization and the message-secret reads behind `DecryptPollVote`/`DecryptSecretEncryptedMessage` survive the canceled login request context while `DeviceFromContext` values are preserved.

Regression test `TestPreparePollWebhookPayloadDecryptsVoteAfterHandlerContextCanceled` decrypts a real stored-secret vote with the handler context already canceled: before the fix it failed with `resolution_status=decrypt_failed` (`failed to get original message secret key: context canceled`), after it resolves normally.

Note: `materializeSecretEditMessage` (pre-existing on main) still decrypts MESSAGE_EDIT envelopes with the raw handler ctx and has the same latent exposure; left untouched here as out of scope for this PR.

`go vet ./...`, `go build ./...`, and `go test ./...` rerun clean on this head.

src/infrastructure/whatsapp/poll_event_test.go
```diff
@@ -0,0 +1,367 @@
...
+func (s *memoryPollStore) AppendPollOption(deviceID, chatJID, pollMessageID string, option domainChatStorage.PollOption) error {
+	definition := s.definitions[pollStoreKey(deviceID, chatJID, pollMessageID)]
+	for _, existing := range definition.Options {
+		if existing.Hash == option.Hash {
+			return nil
+		}
+	}
+	definition.Options = append(definition.Options, option)
+	return nil
+}
```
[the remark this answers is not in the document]
aldinokemal: Fixed in 0e8edf0157838c6d2787982da8687547ee42abff. The stub now returns `poll definition %s not found` for unknown polls, mirroring the real `SQLiteRepository.AppendPollOption` contract (which errors on `sql.ErrNoRows`) rather than the committable suggestion's silent `nil`. Duplicate-hash no-op and append behavior unchanged.

src/infrastructure/whatsapp/poll_event.go
```diff
@@ -0,0 +1,305 @@
...
+		return degraded()
+	}
+	decrypted = utils.UnwrapMessage(decrypted)
+	if kind == "add_option" {
+		return preparePollAddOptionPayload(store, deviceID, chatJID, decrypted.GetPollAddOptionMessage(), pollID)
+	}
+	poll, version := utils.ExtractPollCreationMessage(decrypted)
+	if poll == nil {
+		return degraded()
+	}
```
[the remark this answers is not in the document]
aldinokemal: Fixed in 0e8edf0157838c6d2787982da8687547ee42abff. The add_option branch now falls back to `degraded()` when the decrypted envelope carries no `PollAddOptionMessage`, matching the edit path. Covered by `TestPreparePollWebhookPayloadDegradesAddOptionWithoutInnerMessage`, which encrypts a real POLL_ADD_OPTION envelope around an empty inner message and asserts the degraded payload is still delivered.

src/infrastructure/whatsapp/poll_event.go
```diff
@@ -0,0 +1,305 @@
...
+		if jid := instance.JID(); jid != "" {
+			return jid
+		}
+		return instance.ID()
+	}
+	if client != nil && client.Store != nil && client.Store.ID != nil {
+		return client.Store.ID.ToNonAD().String()
+	}
+	return ""
+}
```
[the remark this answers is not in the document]
aldinokemal: Not changed. `pollDeviceID` intentionally mirrors the exact derivation `CreateMessage` uses to stamp `device_id` on chats/messages (sqlite_repository.go: context instance `JID()` → `ID()` → `client.Store.ID.ToNonAD().String()`), and `CreateMessage` is what defines the partition key poll rows must join against.

Two problems with deriving from `client.Store.ID` only:
1. When the context carries a device instance whose storage wrapper was keyed by the slot id (`ensureInstance` falls back to the slot id when `inst.JID()` is empty), messages are stored under that id — poll definitions written under `Store.ID.ToNonAD()` would then never match lookups or the `DeleteMessageByDevice`/`DeleteChatByDevice`/`DeleteDeviceData` cleanup predicates.
2. `DeviceInstance.JID()` *is* `client.Store.ID.ToNonAD().String()` once paired, so for the normal case the two derivations already agree; the fallback only differs where `CreateMessage` also differs, which is the consistency that matters.

Dropping the context path would desynchronize poll rows from the message rows written for the same event, which is the exact failure mode this comment is trying to prevent.

aldinokemal:
Review feedback resolved in 0e8edf0157838c6d2787982da8687547ee42abff:

- **P1 (canceled handler context)**: poll payload preparation now runs under `context.WithTimeout(context.WithoutCancel(ctx), 30s)`; regression test `TestPreparePollWebhookPayloadDecryptsVoteAfterHandlerContextCanceled` decrypts a real stored-secret vote with the parent context canceled (failed before the fix, passes after).
- **add_option missing inner message**: falls back to `degraded()` like the edit path, with a real-crypto regression test.
- **`memoryPollStore.AppendPollOption` nil deref**: stub now errors on unknown polls, mirroring the real repository contract.
- **testify nitpick**: new tests in `sqlite_repository_poll_test.go` (cleanup scenarios now table-driven), `whatsapp_poll_test.go` (version table kept), and `send_test.go` converted to `assert`/`require`.
- **`pollDeviceID` derivation**: not changed — see reply on that thread; the fallback chain deliberately matches `CreateMessage`'s `device_id` derivation, which is the partition key poll rows must stay consistent with.

`go vet ./...`, `go build ./...`, `go test ./...` all pass on this head.

src/infrastructure/whatsapp/poll_event.go
```diff
@@ -0,0 +1,315 @@
...
+
+func pollChatID(ctx context.Context, client *whatsmeow.Client, evt *events.Message) string {
+	if evt == nil {
+		return ""
+	}
+	return NormalizeJIDFromLID(ctx, evt.Info.Chat, client).ToNonAD().String()
```
aldinokemal: [P1] This makes the persisted lookup key depend on whether the LID mapping exists *at this moment*. `NormalizeJIDFromLID` intentionally falls back to the raw `@lid` when `GetPNForLID` has no mapping. If a poll creation is stored while that mapping is unavailable, its definition is keyed by `123@lid`; after the mapping arrives, the vote for the same chat is looked up under `628…@s.whatsapp.net`, so decryption succeeds but the webhook incorrectly reports `definition_missing`. The inverse can happen for a sent poll stored under PN if a later direct-chat event is only addressable by LID. Please carry stable lookup candidates (raw + normalized, and the target message key's remote JID where applicable), migrate/alias a raw-key hit to the canonical key, and add a regression where mapping availability changes between creation and vote.

src/infrastructure/whatsapp/history_sync.go
```diff
@@ -238,6 +238,13 @@ func processConversationMessages(ctx context.Context, data *waHistorySync.Histor
 				continue
 			}
 
+			if poll, version := utils.ExtractPollCreationMessage(msg.GetMessage()); poll != nil {
+				definition := pollDefinitionFromCreation(deviceID, chatJID, messageID, version, poll)
+				if err := chatStorageRepo.UpsertPollDefinition(definition); err != nil {
```
aldinokemal, line 243: [P1] History replay can roll a poll definition backwards. This always calls `UpsertPollDefinition` for the original creation, and that repository method replaces `question/options_json/selectable_option_count/version` unconditionally. If the persisted definition already contains a newer live poll edit or appended option, a later RECENT/INITIAL_BOOTSTRAP chunk containing the original creation erases those newer options. A subsequent vote then resolves against stale data. Creation replay should be insert-if-missing (or revision/timestamp-ordered), while actual poll edits remain the replace path; if rebuilding from scratch, replay historical edit/add-option events in order. Please add a regression: existing definition + appended option, then replay original creation, and verify the newer option is retained.

aldinokemal (commented):
Verdict: REQUEST CHANGES

The previous canceled-handler-context blocker is fixed on this head and I resolved that thread after verifying the detached 30s context and real stored-secret regression test against the pinned whatsmeow `33cfac511629` behavior.

Two data-integrity blockers remain before this is merge-safe:

1. Poll-definition identity is not stable across late LID→PN mapping. The persistent key is derived from whichever mapping happens to be available for the current event, so the same direct chat can be stored under `@lid` and later looked up under `@s.whatsapp.net` (or vice versa). Decryption can succeed while selection-name resolution incorrectly degrades to `definition_missing`.
2. Replaying an original poll creation from RECENT/INITIAL_BOOTSTRAP history unconditionally overwrites an existing definition. That can roll back options already added/edited by newer live events, which defeats the restart-persistence goal and makes later votes partially resolve against stale options.

Please make poll creation persistence monotonic and make lookup tolerant of raw/normalized chat identity transitions, with regressions for both cases. Current `main` is exactly the PR base (`b78ec206`), and `src/go.mod` pins `go.mau.fi/whatsmeow@33cfac511629`. There are no GitHub Actions runs for this exact head; the PR reports local test/race/vet/build checks passing.

aldinokemal:
Resolved the remaining data-integrity blockers in `2b5d39f`:

- Poll lookup now checks normalized, raw, and referenced message-key chat identities. Live edit/add-option updates reuse the stored definition’s actual chat key. `TestPreparePollWebhookPayloadFindsDefinitionStoredBeforeLIDMapping` exercises real encrypted vote decryption after a late LID→PN mapping.
- Poll-definition writes are now monotonic. Callers preserve WhatsApp event timestamps, and `UpsertPollDefinition` ignores definitions that are not strictly newer. `TestUpsertPollDefinitionDoesNotReplaceNewerDefinition` verifies a replayed history creation cannot roll back a newer edited/extended definition.

Fresh verification on this head:

- `go test ./infrastructure/chatstorage ./infrastructure/whatsapp ./usecase ./pkg/utils`
- `go test -race ./infrastructure/chatstorage ./infrastructure/whatsapp ./usecase`
- `go test -tags purego ./infrastructure/chatstorage ./infrastructure/whatsapp ./usecase`
- `go test ./...`
- `go vet ./...`
- standard and `-tags purego` builds
- `git diff --check`

All passed.

src/infrastructure/whatsapp/history_sync.go
aldinokemal: Fixed in `2b5d39f`. Definition callers now carry WhatsApp event/server timestamps, and `UpsertPollDefinition` updates an existing row only when the incoming timestamp is strictly newer. `TestUpsertPollDefinitionDoesNotReplaceNewerDefinition` writes an edited definition with an added option, replays an older creation, and verifies the newer question/options/timestamp remain intact.

src/infrastructure/whatsapp/poll_event.go
aldinokemal: Fixed in `2b5d39f`. Poll lookup now tries the normalized event chat, raw event chat, and raw/normalized `pollCreationMessageKey.remoteJID`/target-key aliases. Edit and add-option paths reuse the matched definition’s actual persisted `chat_jid` instead of creating a second identity. `TestPreparePollWebhookPayloadFindsDefinitionStoredBeforeLIDMapping` stores the definition under `@lid`, adds the LID→PN mapping, then decrypts and resolves a real vote.

src/infrastructure/whatsapp/history_sync_test.go
```diff
@@ -77,12 +77,73 @@ func TestProcessConversationMessagesPersistsReactionEvents(t *testing.T) {
...
+	if repo.definition == nil || repo.definition.DeviceID != deviceID || repo.definition.PollMessageID != "POLL-HISTORY-1" {
+		t.Fatalf("poll definition not persisted: %+v", repo.definition)
+	}
+	if repo.definition.Question != "History poll" || len(repo.definition.Options) != 2 || repo.definition.Version != "v3" {
+		t.Fatalf("unexpected poll definition: %+v", repo.definition)
+	}
+	wantTimestamp := time.Unix(int64(timestamp), 0)
+	if !repo.definition.UpdatedAt.Equal(wantTimestamp) {
+		t.Fatalf("poll definition updated_at = %s, want history timestamp %s", repo.definition.UpdatedAt, wantTimestamp)
+	}
```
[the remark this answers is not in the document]
aldinokemal: Fixed in `d2ac0ab`. The history timestamp assertion now uses `assert.True` while preserving the `time.Time.Equal` comparison and diagnostic values.

src/infrastructure/whatsapp/poll_event.go
```diff
@@ -0,0 +1,378 @@
...
+	deviceID := pollDeviceID(ctx, client)
+	chatIDs := pollChatIDs(ctx, client, evt)
+	chatJID := ""
+	if len(chatIDs) > 0 {
+		chatJID = chatIDs[0]
+	}
+	msg := utils.UnwrapMessage(evt.Message)
+
+	if poll, version := utils.ExtractPollCreationMessage(msg); poll != nil {
+		definition := pollDefinitionFromCreation(deviceID, chatJID, evt.Info.ID, version, poll, evt.Info.Timestamp)
```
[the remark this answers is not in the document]
aldinokemal: Fixed in `d2ac0ab`. After exact raw/normalized/message-key chat candidates miss, lookup falls back to `(device_id, poll_message_id)`. The repository returns a definition only when that identity is unique and errors if the same poll ID exists in multiple chats. `TestPreparePollWebhookPayloadFindsPNDefinitionForUnmappedLIDVote` covers a real encrypted PN-created poll followed by an unmapped LID-only vote; repository tests cover the ambiguity guard.

src/usecase/send.go
```diff
@@ -1474,10 +1474,47 @@ func (service serviceSend) SendPoll(ctx context.Context, request domainSend.Poll
...
+		ChatJID:               recipient.ToNonAD().String(),
+		PollMessageID:         pollID,
+		Question:              request.Question,
+		Options:               options,
+		SelectableOptionCount: uint32(request.MaxAnswer),
+		Version:               "v1",
+	}
+	if len(sentAt) > 0 {
+		definition.UpdatedAt = sentAt[0]
+	}
```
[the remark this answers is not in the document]
aldinokemal: Fixed in `d2ac0ab`. `ValidateSendPoll` now rejects positive `MaxAnswer` values above `math.MaxUint32` before the existing `len(options)` bound, preventing the narrowing conversion from wrapping. The regression is architecture-aware and skips on 32-bit targets where `int` cannot represent such a value.

aldinokemal (commented):
Verdict: REQUEST CHANGES

Re-review of `d2ac0ab0fb9e72351c3fade423fa0ee7a5e1598f` against current `main` (`b78ec20697da837754bfd24b4156db7fd728a0da`) and the pinned `go.mau.fi/whatsmeow@33cfac511629`.

The previous LID-alias and history-replay blockers are materially addressed: lookup now considers raw/normalized chat identities with an ambiguity-safe device+message-ID fallback, and stale history creations no longer overwrite a strictly newer definition. The canceled-handler-context fix also remains correct.

Two blocking issues remain:

1. Poll-definition ordering still relies on timestamps that are not a safe total order. The pinned whatsmeow parses message `t` through `ag.UnixTime("t")` (second precision), while `AppendPollOption` writes `updated_at = time.Now()` (local wall clock with sub-second precision). A legitimate edit in the same WhatsApp second can therefore compare equal and be dropped, and an add-option can advance `updated_at` ahead of subsequent server timestamps (especially with clock skew), causing later live edits to be ignored. This can leave stale options/question data and produce incorrect/partial vote resolution.
2. Poll preprocessing now performs LID normalization/store I/O synchronously for every incoming message before establishing that the message is a poll. In pinned whatsmeow, `dispatchEvent` invokes registered handlers synchronously; therefore ordinary LID chat traffic can now block event dispatch on `GetPNForLID`/SQLite contention for up to this 30s budget. Poll-specific work needs a cheap type gate before any normalization/store/crypto work.

Please add regression coverage for same-second poll updates / local-clock skew, and for a normal non-poll LID message taking the no-I/O fast path. There are still no GitHub Actions runs associated with this exact head; the PR reports local test/vet/build success.

src/infrastructure/chatstorage/sqlite_repository.go
```diff
@@ -730,12 +739,189 @@ func (r *SQLiteRepository) DeleteMessageByDevice(deviceID, id, chatJID string) e
...
+	defer tx.Rollback()
+
+	var existingUpdatedAt time.Time
+	err = tx.QueryRow(`
+		SELECT updated_at FROM poll_definitions
+		WHERE device_id = ? AND chat_jid = ? AND poll_message_id = ?
+	`, definition.DeviceID, definition.ChatJID, definition.PollMessageID).Scan(&existingUpdatedAt)
+	switch {
+	case err == nil:
+		if !definition.UpdatedAt.After(existingUpdatedAt) {
```
aldinokemal, line 786: **P1 — protocol timestamps are not a total order here.** `definition.UpdatedAt` for live/history poll events comes from `events.Message.Info.Timestamp`; in pinned whatsmeow `33cfac511629`, `parseMessageInfo` sets that from `ag.UnixTime("t")`, so distinct poll updates can legitimately have the same second-level timestamp. This strict `After` then drops the later live update. There is a second ordering problem in `AppendPollOption`: it writes `updated_at = time.Now()`, mixing local wall-clock/sub-second time with WhatsApp server timestamps. Example: creation at `10:00:00`, add-option received locally at `10:00:00.900`, then a valid edit whose WhatsApp `t` is still `10:00:00`; the edit is silently ignored. Clock skew can extend that suppression well beyond one second. Please use an ordering model that distinguishes stale history replay from live arrival without mixing local/server clocks (e.g. protocol timestamp plus a monotonic/revision/source rule), and add tests for same-second creation→edit and add-option→edit under a clock-ahead scenario.

src/infrastructure/whatsapp/poll_event.go
```diff
@@ -0,0 +1,384 @@
...
+		return nil
+	}
+	// The event handler runs with the context captured at registration; for
+	// REST-initiated logins that is the HTTP request context, canceled once
+	// the request returns. Detach so LID normalization and message-secret
+	// reads during vote/edit decryption survive, keeping device-scoped values.
+	ctx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 30*time.Second)
+	defer cancel()
+	deviceID := pollDeviceID(ctx, client)
+	chatIDs := pollChatIDs(ctx, client, evt)
```
aldinokemal, line 222: **P1 — this puts store I/O on every LID message before we know it is a poll.** `preparePollWebhookPayload` is called synchronously at the top of `handleMessage`, and `pollChatIDs` immediately calls `NormalizeJIDFromLID`, which does `client.Store.LIDs.GetPNForLID`. In pinned whatsmeow, `dispatchEvent` runs registered handlers synchronously, so an ordinary text/image/reaction from an `@lid` chat can now stall WhatsApp event dispatch on SQLite/store contention for up to the 30s context even though poll processing is irrelevant. Please unwrap/classify the message first and return `nil` for non-poll types before creating the detached context, resolving device/chat aliases, or touching crypto/storage. Add a regression test proving a normal LID message takes that no-poll fast path.

PR #820, opened 2026-09-01
fix(message,send): quoted replies show pre-edit text and an unresolvable participant

alenphilipgeorge:
## Context

Two independent bugs that both surface through `mergeReplyContext`, which builds a reply's quoted context out of the locally stored copy of the message being quoted:

```go
contextInfo.StanzaID     = replyMessageID
contextInfo.Participant  = proto.String(message.Sender)
contextInfo.QuotedMessage = &waE2E.Message{Conversation: proto.String(message.Content)}
```

For a message sent through the API, **both** of those fields can be wrong.

### 1. `UpdateMessage` never writes the edit to chat storage

`UpdateMessage` sends the edit to WhatsApp and returns. Nothing updates the stored copy, so it keeps the text as first sent, indefinitely.

An edit that *arrives* from another device is already handled — `CreateMessage` spots the `MESSAGE_EDIT` protocol message and calls `storeEditedMessage`. An edit we send ourselves is never echoed back to us, so nothing applies it.

Measured on a live instance before the change: of **461** recorded edits, 425 inbound and 36 own, **every one was event-sourced and none came from the API**.

### 2. Sent messages store a device-suffixed sender

`StoreSentMessageWithContext` is handed `client.Store.ID.String()`, which keeps the device suffix (`6281234567890:32@s.whatsapp.net`) — a linked companion is never device 0.

Every other path already normalises: `CreateMessage` stores `normalizedSender.ToNonAD()`, and history sync and `MarkAsRead` both use `client.Store.ID.ToNonAD()`. So sent messages end up the only rows whose sender is not a plain user JID, and that value goes straight into `ContextInfo.Participant`, where it matches no participant of the chat.

Measured on the same instance: **162 of 41,899** stored messages carried the suffix, and every one was sent through the API. None of the 41,737 event-sourced rows did.

### What a recipient sees

Send a message, edit it, then reply to it. The bubble shows the edited text — the edit itself is fine, it travels over the wire and never touches storage. But the **quote** in the reply shows the *original* text, with a participant the client cannot resolve. If the edit changed the opening line, the quote and the message above it share no visible text, and the quote reads as an unrelated message.

## Changes

- `fix(send)`: new `whatsapp.OwnSenderJID()`, used by both sent-message paths (`usecase/send.go`, `infrastructure/whatsapp/auto_reply.go`) so the derivation lives in one place.
- `fix(message)`: `UpdateMessage` mirrors the edit into chat storage via a new `updateStoredMessage`, following the same shape as the existing `deleteStoredMessage`. The edit is replayed as the same `events.Message` an inbound edit produces, so it runs through the existing tested edit path (content update + `message_edits` row) instead of a second implementation that could drift.
- `UpdateMessage` now also goes through `service.validateJID` / `service.sendMessage`, as `RevokeMessage` and `DeleteMessage` already do. Both seams default to exactly what it called before, so behaviour is unchanged — it is what makes the method reachable from the existing test harness.

Happy to split this into two PRs if you'd prefer them separate.

## Test Results

**Unit tests** — `go test ./...` → all 13 packages pass, no failures.

New tests:
- `TestUpdateMessageSyncsEditToChatStorage`
- `TestUpdateMessageRecordsEditHistory`
- `TestUpdateMessageLeavesOtherDevicesAlone`
- `TestUpdateMessageKeepsStoredContentWhenWhatsAppFails` — mirrors the existing `TestMessageActionsKeepStoredMessageWhenWhatsAppFails`: storage must not drift ahead of the recipient when the edit never reached WhatsApp
- `TestOwnSenderJIDStripsTheDeviceSuffix` / `LeavesAPlainJIDAlone` / `IsEmptyWhenUnavailable`

Mutation-checked: with the `updateStoredMessage` call removed, the tests fail (`stored content` instead of `edited content`, and 0 history rows), so they do catch the bug.

**Live instance** — patched build, real paired session, real group. Same account and code path, before vs after:

| | `sender` stored | `content` stored |
|---|---|---|
| message edited **before** the fix | `9193xxxxxxxx:32@s.whatsapp.net` | `hiiiii` — stale, actual text is `hiiiii there` |
| message edited **after** the fix | `9193xxxxxxxx@s.whatsapp.net` | `ENG2775 clean test: EDITED first line` — current |

After the fix the edit lands in storage with `updated_at` bumped and a `message_edits` row holding the correct previous/new content, and a reply to that message quotes the edited text on the recipient's device.

Note the fix applies to edits made *after* it ships; copies already stale in an existing store stay stale, since there is nothing left to replay them from.


src/usecase/message.go
```diff
@@ -389,6 +391,70 @@ func (service serviceMessage) DeleteMessage(ctx context.Context, request domainM
...
+					ID:        proto.String(messageID),
+					RemoteJID: proto.String(recipient.String()),
+					FromMe:    proto.Bool(true),
+				},
+				EditedMessage: &waE2E.Message{Conversation: proto.String(newContent)},
+			},
+		},
+	}
+
+	if err := service.chatStorageRepo.CreateMessage(ctx, evt); err != nil {
```
aldinokemal, line 458: 🔴 **P1 — a pending original-message store can overwrite this edit.** `wrapSendMessage` returns after spawning `StoreSentMessageWithContext` in a goroutine. If the caller edits quickly (or the original store is delayed by SQLite contention), this `CreateMessage` can write the edited content first; afterward the older goroutine reaches `StoreSentMessageWithContext -> StoreMessage`, whose existing-row path unconditionally updates `content` back to the original text. That recreates the stale-quote bug this PR is meant to fix. Please make the local write ordering durable so an initial sent-message persistence cannot overwrite a later edit, and add a deterministic regression test that blocks the original store until after `UpdateMessage` has completed.

src/infrastructure/whatsapp/own_sender.go
```diff
@@ -0,0 +1,24 @@
...
+//
+// Every other path already normalises: CreateMessage stores
+// normalizedSender.ToNonAD(), history sync uses client.Store.ID.ToNonAD(), and
+// MarkAsRead uses client.Store.ID.ToNonAD(). This keeps the two sent-message
+// paths consistent with them.
+func OwnSenderJID(client *whatsmeow.Client) string {
+	if client == nil || client.Store == nil || client.Store.ID == nil {
+		return ""
+	}
+	return client.Store.ID.ToNonAD().String()
```
aldinokemal, line 23: 🟠 **Major — this fixes only future writes, while the quote reader still trusts legacy persisted values.** The PR itself found existing rows with senders like `628…:32@s.whatsapp.net`; after upgrade those rows remain unchanged, and `mergeReplyContext` still copies `message.Sender` directly into `ContextInfo.Participant`, so replies to pre-upgrade messages are still unresolvable. Since this changes the persisted sender representation, make the read path backward-compatible (parse/normalize the stored sender before building reply context, including LID handling where appropriate) or migrate the old rows. Please add a regression test that quotes a legacy device-suffixed stored sender after upgrade.

aldinokemal (changes requested):
Reviewed the current head against current `main` (the PR base is exactly current main), the chat-storage write/update paths, reply-context construction, device scoping, existing edit-history implementation, and whatsmeow pinned by `src/go.mod` (`0fadda796019`). The wire-side change is compatible with that pinned revision: `BuildEdit` still builds the edit protocol message there, and `JID.ToNonAD()` is the correct way to strip AD/device identity for the stored own sender.

I found two blocking correctness/backward-compatibility gaps. The first is an ordering race with the existing asynchronous sent-message persistence: a fast edit can be stored first and then overwritten by the older original-content write. The second is that normalizing only new writes leaves every legacy device-suffixed sender already in chat storage broken when `mergeReplyContext` quotes it.

One protocol-detail note while fixing/tests are being updated: pinned whatsmeow unwraps incoming `MESSAGE_EDIT` events before dispatch (`evt.Info.ID` becomes the target ID and `evt.Message` becomes the edited content), so the synthetic direct `ProtocolMessage` used here is a useful local-storage adapter, but it is not literally the same `events.Message` shape an event handler receives at this revision. Please avoid relying on that equivalence in tests/comments.

No PR-triggered GitHub Actions run exists for this exact head; the only commit status currently present is CodeRabbit success. Please run the normal Go test/vet/build checks after the ordering and legacy-row fixes.

src/usecase/message.go
alenphilipgeorge: Confirmed, and fixed in c99edf0.

`StoreMessage`'s existing-row path is `UPDATE messages SET sender = ?, content = ?, ... updated_at = ?` with no guard, so a late original write does roll the edit back.

`StoreSentMessageWithContext` now reads edit history and keeps the stored content when any exists — content only ever moves forward, while the rest of the sent-message metadata is written as before. Placed there rather than in `StoreMessage` so no other caller's semantics change.

Regression tests in `sqlite_repository_sent_edit_order_test.go` drive the writes in the order the race produces rather than depending on the scheduler:

- `TestDelayedSentStoreDoesNotOverwriteAnEarlierEdit` — edit lands first (row does not exist yet), then the delayed original store arrives; content must stay edited
- `TestEditAfterTheSentStoreStillApplies` — ordinary order still works
- `TestSentStoreWritesContentWhenNoEditExists` — the guard must not freeze content for every re-stored message

Mutation-checked: with the guard removed, only the first test fails.

On the protocol note — you are right that the shapes are not identical, and I have removed that claim from the comment. One clarification for the record: at `0fadda796019` the `MESSAGE_EDIT` unwrap is in `ParseWebMessage` (`client.go:1016`), which is the history-sync/web path. The live receive path dispatches `evt.UnwrapRaw()` (`message.go:306` and `:1148`), and `UnwrapRaw` handles `DeviceSentMessage`/`EphemeralMessage`/`ViewOnce` only, not edits — so a live inbound edit still arrives with the `ProtocolMessage` intact, which is what `extractEditedMessage` reads. The comment now describes the replay as a local-storage adapter and names the `ParseWebMessage` difference explicitly.

`go build ./...`, `go vet ./...` and `go test ./...` all pass (13 packages, 0 failures).

src/infrastructure/whatsapp/own_sender.go
alenphilipgeorge: Confirmed, and fixed in cdf2a10 — on the read path, so existing rows are covered without a migration.

`mergeReplyContext` no longer trusts the persisted value. It parses, LID-normalizes and strips AD identity before building the participant, mirroring `RevokeMessage`, which already does exactly this to a stored sender ("Stored senders can still be @lid; whatsmeow's Revoke needs the phone-number form").

One thing worth flagging that the tests turned up: `utils.ParseJID` does **not** report failure on a malformed value — it appends the default server, so `"not-a-jid"` becomes `"not-a-jid@s.whatsapp.net"`. My first version therefore rewrote an unusable value into a *different* unusable value while looking like it passed it through. It now only touches a value that already carries a server.

Tests in `send_test.go`:

- `TestMergeReplyContextNormalizesALegacyDeviceSuffixedSender` — a stored `628123456789:32@s.whatsapp.net` yields participant `628123456789@s.whatsapp.net`
- `TestMergeReplyContextPassesThroughAnUnparseableSender` — a value with no server survives verbatim

Mutation-checked: reverting the call makes the first fail.

Verified against a live instance carrying pre-upgrade rows. The 162 stored senders with a device suffix are unchanged on disk, as expected, and a reply quoting one now builds a normalized participant.

I kept this on the read path rather than migrating the rows, since it also covers anything written by an older binary against the same store. Happy to add a migration as well if you would rather converge the persisted values.

src/infrastructure/chatstorage/sqlite_repository.go
```diff
@@ -2461,6 +2461,27 @@ func (r *SQLiteRepository) StoreSentMessageWithContext(ctx context.Context, mess
...
 		FileLength:    fileLength,
 	}
+	// wrapSendMessage persists asynchronously, so an edit sent moments later can
+	// reach storage BEFORE this does. StoreMessage's existing-row path updates
+	// content unconditionally, so writing the original text now would roll that
+	// edit back — and mergeReplyContext would then quote the rolled-back text,
+	// which is the stale quote the edit sync exists to prevent. Content only ever
+	// moves forward: once an edit is recorded, keep what is stored and let the
+	// rest of the sent-message metadata through.
+	edits, editErr := r.GetMessageEdits(messageID, deviceID)
```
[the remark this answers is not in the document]
alenphilipgeorge: Confirmed and fixed.

The check-then-act was real: an edit committing between `GetMessageEdits` and `StoreMessage` was overwritten by the original content — the exact failure the guard exists to prevent. And the window is wider than statement adjacency suggests, since `busy_timeout` is 30s and history-sync batches hold the writer.

Fixed with one conditional statement rather than a transaction, so no `StoreMessage` plumbing changes and no other caller's semantics move. The check is now an `EXISTS` inside the `UPDATE`:

```sql
content = CASE WHEN EXISTS (
    SELECT 1 FROM message_edits
    WHERE original_message_id = messages.id
      AND chat_jid = messages.chat_jid
      AND device_id = messages.device_id
) THEN content ELSE ? END
```

Everything except content is written as normal; only content is pinned once an edit exists for the row. Device- and chat-scoped, as you asked.

The insert path had the same shape one level down — `UPDATE` affects 0 rows, an edit creates the row, then the `INSERT` conflicts. It now retries the guarded update once, so the edit keeps its content and the sent-message metadata still lands, instead of surfacing a constraint error.

Tests, in `sqlite_repository_sent_edit_order_test.go`:

- `TestConcurrentSentStoreAndEditNeverYieldsOriginalContent` — the interleaving test you asked for: 40 rounds, both writers released from a shared barrier, asserting that once an edit is *recorded* the stored content is never the original
- the three deterministic ordering tests alongside it

Mutation-checked: reverting the `CASE` to an unconditional `content = ?` fails **both** the deterministic and the concurrent test.

One thing worth being explicit about: the concurrent test is a smoke test, not the guarantee. A timing test can always pass by luck. The actual guarantee is structural — the check and the write are a single statement, so there is no interleaving point between them.

`go build ./...`, `go vet ./...`, `go test ./...` (and `-race` on this package) all pass — 13 packages, 0 failures.

aldinokemal (approved):
Re-reviewed current head after the requested fixes. Both owner blockers are addressed: sent-message persistence now uses an atomic guarded UPDATE keyed by message/chat/device so a delayed original write cannot roll back a committed edit, including the row-creation race; and reply construction now normalizes legacy persisted AD/device-suffixed senders on the read path, including the existing LID→PN normalization path, before ToNonAD.

I also rechecked the protocol-facing pieces against the pinned `go.mau.fi/whatsmeow@0fadda796019`: `BuildEdit` still emits `ProtocolMessage_MESSAGE_EDIT`, `SendMessage` returns a distinct sent-message ID/timestamp after server ack, and `JID.ToNonAD()` removes device/agent identity as expected. Device scoping remains intact and the new storage tests cover delayed-store, normal ordering, cross-device isolation, edit history, remote failure, and concurrent interleavings.

No blocking correctness/protocol issue remains. Note that GitHub currently has no PR-triggered workflow run for this exact head, so the author-reported `go test ./...`, `go vet ./...`, build, and race checks are not independently represented in Actions.

alenphilipgeorge:
@aldinokemal all three review items are addressed and pushed:

- **P1 ordering race** — `StoreSentMessageWithContext` no longer lets a delayed original write roll back an edit, and after CodeRabbit's follow-up the check is an `EXISTS` inside the `UPDATE`, so check and write are a single statement with no interleaving point. Deterministic + concurrent tests, both failing when the guard is reverted.
- **Legacy senders** — normalized on the read path in `mergeReplyContext`, so pre-upgrade rows are covered without a migration.
- **Protocol note** — comment corrected; it no longer claims equivalence with a dispatched event.

`go build ./...`, `go vet ./...`, `go test ./...` all pass (13 packages, 0 failures), plus `-race` on the touched package.

**Why this one matters to us, and why a tagged release would help.**

We hit this in production. An operator sent a message, corrected it 32 seconds later, and replied to it hours afterwards. The recipient's quote showed the *pre-edit* text — wording the message no longer contained — with an unresolvable participant. From their side it read as a reply to a message that did not exist, and they asked what it referred to. We had to tell a customer it was a glitch.

Two things make it hard to work around:

1. **There is no version to pin to.** The behaviour is identical on the tag we run (v8.11.0) and on current `main`, so upgrading does not avoid it.
2. **It cannot be fixed from the client side.** The quote is built from GOWA's own stored copy, and the send API exposes only `reply_message_id` — no way to supply the quote text or participant. We verified the chat REST surface is read-only for messages, so we cannot repair the stored row either.

It is also silent: the edit reaches WhatsApp correctly and every operator-facing surface looks right, so nothing indicates the recipient saw something different.

So a merge plus a tagged release would let us move off a local build and back onto the published image — we consume `aldinokemal2104/go-whatsapp-web-multidevice` directly and would rather not maintain a fork for this.

Thanks for looking at it, and for the quick, detailed review.

aldinokemal (approved):
Re-reviewed the exact current head `333472bc276592eb600f349d2361d7d03b5dc406` after the branch was brought up to current `main`. Since the previous owner approval, the only branch delta is the current-main dependency refresh in `src/go.mod`/`src/go.sum`; the feature diff itself remains the same nine files.

I rechecked the protocol-sensitive pieces against the authoritative current pin, `go.mau.fi/whatsmeow@de26b4ab6499`. Compared with the previously reviewed `0fadda796019`, the upstream changes are protobuf/download/client-payload updates; the relevant `BuildEdit` construction and `JID.ToNonAD()` semantics are unchanged. `BuildEdit` still uses `MESSAGE_EDIT` with the original message key, and `ToNonAD()` still removes AD/device identity while retaining the stable user/server identity.

The two earlier correctness concerns remain resolved: the delayed sent-message persistence uses a device/chat/message-scoped atomic guarded UPDATE plus insert-race retry, so a committed edit cannot be rolled back by the older original write; and reply construction normalizes legacy persisted AD/LID senders on the read path before using them as the quoted participant. Device isolation remains intact.

No new blocking correctness, protocol, storage, race, or API-contract issue was introduced by the main merge. GitHub still has no PR-triggered Actions run for this exact head; the visible commit status is CodeRabbit success, so normal repository CI should still run before merge.

alenphilipgeorge:
@aldinokemal Hey bro, Can you test and merge this to main and release it. This is really important for the product we are building.

PR #822, opened 2026-09-02
fix(chatwoot): don't reopen resolved conversations on history sync

aguinaldotupy:
## Context

History sync reopened conversations that agents had already resolved.

Both sync paths resolved the Chatwoot conversation before checking whether the chat had anything new to post. `FindOrCreateConversation` reopens a resolved conversation when `ChatwootReopenConversation` is set, and history auto-sync latches once per device per process — so every restart reopened every resolved thread in the inbox, even when dedupe worked and no message was created.

Three call sites:

- **`syncChat`** — resolved contact and conversation as steps 1 and 2, before reading the messages.
- **`syncChatPG`** — called `ImportChat` unconditionally. Per-message idempotency inside the importer cannot prevent the reopen: `pgimport`'s own `findOrCreateConversation` reopens the thread before it gets to skipping the duplicate rows.
- **REST media pre-pass** — resolved a conversation whenever the chat held downloadable media, even when all of it was already imported.

All three now drop the already-linked messages first and resolve a conversation only when something is left to post. A chat with no messages inside the time window no longer creates one either — that early return used to happen after the conversation had been created.

Already-linked messages still count as synced, so operator-facing totals are unchanged. `ImportChat` now receives only the pending messages, so its `wrote`/`skipped` totals do not double-count them.

## Test Results

`cd src && go vet ./...` clean. `go test ./...` — 13/13:

```
ok  	.../cmd                                 0.016s
ok  	.../infrastructure/chatstorage          0.362s
ok  	.../infrastructure/chatwoot            10.071s
ok  	.../infrastructure/chatwoot/pgimport    0.027s
ok  	.../infrastructure/uiasset              0.018s
ok  	.../infrastructure/whatsapp             1.418s
ok  	.../pkg/utils                          21.154s
ok  	.../ui/mcp                              0.030s
ok  	.../ui/mcp/oauth                        0.056s
ok  	.../ui/rest                             0.021s
ok  	.../ui/rest/middleware                  0.072s
ok  	.../usecase                             0.253s
ok  	.../validations                         0.034s
```

New unit tests in `sync_reopen_test.go` cover both paths and the media pre-pass. Each was confirmed to fail with the fix reverted.

For the Postgres path I brought up a Chatwoot container (`pgvector/pgvector:pg17` plus `rails db:chatwoot_prepare`) and ran `syncChatPG` against the real schema, rather than more sqlmock. `sync_pg_integration_test.go` is gated behind `CHATWOOT_TEST_DB_URI` and skips when it is unset, so `go test ./...` and CI are unaffected; `go.mod`/`go.sum` are untouched. The setup is in the test file's header comment.

```
--- PASS: TestSyncChatPGIntegration_ImportIsIdempotent (0.06s)
--- PASS: TestSyncChatPGIntegration_ResolvedConversationStaysResolved (0.04s)
--- PASS: TestSyncChatPGIntegration_NewMessageStillImportsAndReopens (0.04s)
ok  	.../infrastructure/chatwoot	0.146s
```

`ResolvedConversationStaysResolved` is the regression test. Against the same schema, before the `syncChatPG` fix:

```
conversation status = 0, want 1 (resolved);
the re-import reopened a thread with nothing new to add
```

`NewMessageStillImportsAndReopens` guards the other direction: a new message must still import and reopen the thread.

`go test -race ./infrastructure/chatwoot/...` passes with and without the integration tests enabled.

---

https://claude.ai/code/session_019NEWA4g15o2wTm1jy9Mgct


aldinokemal (dismissed):
Reviewed exact head `4dcedb5d181ca5eceb7244c29f9fef8393d012b6` against current `main` (`b2a57a2062d5dcebcb70c6dfd05c8aec5b613868`), which is also this PR's exact base. I traced both Chatwoot history-sync paths, the REST media pre-pass, local message-link persistence, progress accounting, direct pgimport behavior, and the new regression/integration tests.

The ordering change is correct: `syncChat` and `syncChatPG` now read the time-windowed messages and remove messages with an already-persisted nonzero Chatwoot link before resolving a contact/conversation, so a history run with nothing new no longer calls `FindOrCreateConversation` and therefore cannot reopen a resolved thread just because auto-sync ran again. Pending messages still resolve the conversation and follow the existing send/import paths; already-linked rows still count as synced, while pgimport wrote/skipped/failed totals are applied only to the subset actually passed to the importer. The media pre-pass applies the same guard before its only conversation-resolution side effect.

I also checked that the new prefilter deliberately matches the existing `syncMessageWithOptions` link semantics (device + WhatsApp message ID, nonzero Chatwoot message ID), so it does not introduce a new dedupe contract or cross-layer state mutation. Direct-Postgres import remains restricted to the legacy/global config as on current main, and this PR does not change whatsmeow/protocol behavior, device/session lifecycle, migrations, REST/MCP/websocket contracts, or Chatwoot retry semantics.

The new tests cover the important directions: all-linked and empty REST chats do not touch Chatwoot, a pending message still resolves a conversation, the media pre-pass skips linked/text-only rows but runs for pending media, and the optional real-Postgres tests exercise idempotency plus resolved→resolved and new-message→reopened behavior against a Chatwoot schema. No blocking correctness, data-loss, race, or compatibility issue found.

Non-blocking: CodeRabbit's open integration-test comment about replacing bare `Exec`/`QueryRow` with context-aware variants is reasonable test hygiene and would satisfy the configured `noctx` lint, but it does not affect production behavior. There is no PR-triggered GitHub Actions run attached to this exact head; the author-reported `go test ./...`, `go vet ./...`, race tests, and real-Postgres integration run are therefore not independently represented in Actions.

src/infrastructure/chatwoot/sync_pg_integration_test.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+	return db, dsn
+}
+
+// truncateChatwootData clears the rows an import writes while leaving the
+// account, inbox and agent fixture in place, so each test starts from a known
+// empty inbox without re-running the migrations.
+func truncateChatwootData(t *testing.T, db *sql.DB) {
+	t.Helper()
+	if _, err := db.Exec(`TRUNCATE messages, conversations, contact_inboxes, contacts RESTART IDENTITY CASCADE`); err != nil {
```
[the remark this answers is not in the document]
aguinaldotupy: Addressed in `baa61f9`. A 30s test-scoped context now threads through the helpers and the `syncChatPG` calls, with `ExecContext`/`QueryRowContext` throughout — an unreachable database now fails in 0.00s naming the ping rather than blocking until the global test timeout.

Also renamed `conversationStatusByContactCount` to `countConversationsAndMessages`; it reports row counts, not a status, and nothing about it was per-contact.

Test-only — `sync.go` is untouched by this commit.

src/infrastructure/chatwoot/sync.go
```diff
@@ -470,6 +483,35 @@ func chatwootRESTMediaCandidates(messages []*domainChatStorage.Message, opts Syn
...
+func (s *SyncService) splitAlreadyLinkedMessages(messages []*domainChatStorage.Message) (pending []*domainChatStorage.Message, alreadyLinked int) {
+	pending = make([]*domainChatStorage.Message, 0, len(messages))
+	for _, msg := range messages {
+		if msg == nil {
+			continue
+		}
+		if msg.ID != "" && msg.DeviceID != "" && s.chatStorageRepo != nil {
+			existing, err := s.chatStorageRepo.GetChatwootMessageLinkByWhatsAppID(msg.DeviceID, msg.ID)
+			if err != nil {
+				logrus.Warnf("Chatwoot Sync: Failed to look up message link for %s: %v", msg.ID, err)
```
aldinokemal: **Blocker — a transient link lookup failure can still reopen a resolved conversation with no new message.** This error path deliberately keeps the message in `pending`, so `syncChat`/`syncChatPG` can immediately call `findOrCreateHistoryConversation` and reopen the thread. Then `syncMessageWithOptions` repeats the same lookup; if the transient error has cleared, it finds the existing link and skips the message. Net result: zero new messages, but the conversation was reopened — exactly the regression this PR is meant to prevent. Please propagate/retry this lookup error and abort before any Chatwoot call when dedupe state is unknown. Add a test where the first lookup errors and the next would return an existing link, asserting that Chatwoot receives no request.

aldinokemal (changes requested):
Re-reviewed the current head after the test-only follow-up. The `ExecContext`/`QueryRowContext` change is good and does not alter production behavior, and the branch is still based exactly on current `main`.

I found one remaining correctness blocker in the new pre-dedupe flow. The purpose of this PR is to avoid calling `FindOrCreateConversation` unless there is actually something new to post, because that call can reopen a resolved conversation. However, `splitAlreadyLinkedMessages` treats a link-lookup error as "pending" and continues. That means a transient chatstorage read error can still trigger the exact regression: first lookup errors -> the message is considered pending -> `findOrCreateHistoryConversation` reopens the thread -> `syncMessageWithOptions` retries the lookup, sees the existing link, and skips the message. The conversation is reopened even though nothing new was posted.

Please make dedupe-state lookup failures fail closed before any Chatwoot side effect (for example, return an error from the partition helper and abort/retry the chat), and add a regression test where the first link lookup fails but a subsequent lookup would show the message is already linked. The expected result should be no contact/conversation request at all.

There are no pull-request-triggered GitHub Actions runs for this exact head; the only exposed commit status is CodeRabbit success. The PR's local `go vet`, `go test`, race, and real-Postgres results are useful, but repository CI has not independently verified them.

aguinaldotupy:
Good catch — you are right, and the failure mode is exactly as you traced it. Fixed in `1fe86e0`.

`splitAlreadyLinkedMessages` now returns an error and aborts on the first failed lookup instead of treating the message as pending. `syncChat` and `syncChatPG` propagate it before any Chatwoot call, so the chat counts as failed and the next sync retries it. The REST media pre-pass skips itself with a warning rather than aborting the import: at that point there is already something pending, so the conversation is going to be resolved regardless, and without a readable link state the pre-pass cannot tell whether the media is already imported.

Added `TestSyncChatAbortsWhenLinkLookupFails` with the repo you described — the first lookup errors, any later one would report the message as already linked. It asserts zero Chatwoot requests, and that the partition stops at the first failure rather than probing on. I confirmed it reproduces the regression: reverting to the fail-open branch makes it fail with `Chatwoot received 3 requests, want 0` (the three are the `FindOrCreateContact` retries), so the reopen path was genuinely reachable.

`go vet ./...`, `go test ./...` (13/13), `-race`, and the gated real-Postgres tests are all green.

On CI: agreed, and I have no way to address it from here — the repository has no PR-triggered workflow that runs `go test` or `go vet`, so the results above remain author-reported.

Sorry for dismissing your approval twice. This one was a correctness blocker rather than hygiene, so it had to go in.

src/infrastructure/chatwoot/sync.go
```diff
@@ -470,6 +489,39 @@ func chatwootRESTMediaCandidates(messages []*domainChatStorage.Message, opts Syn
...
+func (s *SyncService) splitAlreadyLinkedMessages(messages []*domainChatStorage.Message) (pending []*domainChatStorage.Message, alreadyLinked int, err error) {
+	pending = make([]*domainChatStorage.Message, 0, len(messages))
+	for _, msg := range messages {
+		if msg == nil {
+			continue
+		}
+		if msg.ID != "" && msg.DeviceID != "" && s.chatStorageRepo != nil {
+			existing, lookupErr := s.chatStorageRepo.GetChatwootMessageLinkByWhatsAppID(msg.DeviceID, msg.ID)
+			if lookupErr != nil {
+				return nil, 0, fmt.Errorf("failed to look up message link for %s: %w", msg.ID, lookupErr)
```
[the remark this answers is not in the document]
aguinaldotupy: Agreed and fixed in `4132a8e`.

The decisive argument is internal consistency: the partition already treats a linked row as synced because it is present in Chatwoot, and a lookup that fails on a *later* message says nothing about the ones resolved before it. Discarding the count contradicted the rule the same function documents.

`splitAlreadyLinkedMessages` now returns the partial count alongside the error, and both callers record it before propagating. The doc comment states the contract explicitly: on error `pending` is nil and must not be used, while the count stays valid.

There is a second effect worth naming: `progress.AddMessages(len(messages))` runs before the partition, so aborting without recording the count left those messages counted in the total but in neither synced nor failed. This closes part of that gap.

Added `TestSyncChatKeepsLinkedProgressWhenALaterLookupFails`, which fails the lookup for one specific message ID so a confirmed link can sit before the failure. It asserts `SyncedMessages == 1` and zero Chatwoot requests. Reverting just the discarded count makes it fail with `SyncedMessages = 0, want 1`.

`go vet ./...`, `go test ./...` (13/13), `-race`, and the gated real-Postgres tests are green.

src/infrastructure/chatwoot/sync.go
```diff
@@ -419,24 +444,18 @@ func (s *SyncService) syncChatPG(
...
-			s.syncHybridMediaMessagesREST(ctx, conversation.ID, mediaMessages, waClient, opts, isGroup)
-		}
-	}
+	s.restMediaPrePass(ctx, chat, pending, waClient, opts, isGroup)
 
 	result, err := importer.ImportChat(ctx, pgimport.ImportChatRequest{
 		ChatJID:  chat.JID,
 		ChatName: chatName,
-		Messages: messages,
+		Messages: pending,
```
aldinokemal, line 465: 🔴 **P1 — local link state is not durable enough to guard pgimport's reopen side effect.** `ImportChat` commits the Chatwoot transaction before `storeChatwootImportLinks` persists the returned links locally. If the process crashes (or the local link write fails) in that window, Chatwoot already contains this `source_id` but this prefilter sees no local link on the next run, so the message reaches `ImportChat` as `pending`. The importer currently calls `findOrCreateConversation` *before* `insertMessageSavepoint`; with `CHATWOOT_REOPEN_CONVERSATION` enabled it flips a resolved conversation open, and only afterward discovers the existing `source_id` and counts the message as skipped. That recreates the core bug: no new message, but the thread reopens. Please defer reopening an existing resolved conversation until at least one message is actually written (or perform Chatwoot-side dedupe before reopening), and cover the crash/missing-local-link replay case in the pgimport integration test.

aldinokemal (changes requested):
Re-reviewed the current head after the transient-link-lookup fix. The previous owner blocker is addressed: `splitAlreadyLinkedMessages` now fails closed on lookup errors, preserves already-confirmed progress, and the regression tests verify that no Chatwoot request occurs when dedupe state is unreadable. I resolved that old thread.

One blocking durability gap remains on the direct-Postgres path. The new prefilter treats the local `chatwoot_message_links` table as the source of truth for whether `ImportChat` is safe to enter, but the Chatwoot transaction commits before those local links are persisted. A crash or local chatstorage failure after `ImportChat` commits and before `storeChatwootImportLinks` completes leaves the Chatwoot message present while the local link is absent. On the next history sync this code marks the row pending; `pgimport.ImportChat` calls `findOrCreateConversation` first and reopens a resolved conversation, then its source_id dedupe skips the already-existing message. The result is again zero new messages with a reopened conversation.

Please make pgimport's reopen decision depend on an actual write (for example, reuse a resolved conversation ID without changing status, insert/dedupe messages, and only reopen when `MessagesWrote > 0`), or otherwise make the Chatwoot-side idempotency check happen before the reopen side effect. Add a regression test that imports a message into Chatwoot, leaves/removes the corresponding local link, resolves the conversation, reruns `syncChatPG`, and verifies the conversation stays resolved when the importer only skips the existing source_id.

The PR still bases exactly on current `main` and is mergeable. There are no GitHub Actions workflow runs attached to head `4132a8ef354c554318c688b1b36a177bc085a210`; the author-reported test/vet/race results are therefore not independently represented in Actions.

src/infrastructure/chatwoot/sync.go
aguinaldotupy: Fixed in `7a96d5b`, and the review turned up that the same gap existed on the REST side, so this commit closes both.

**pgimport (your P1).** `findOrCreateConversation` no longer changes status; it reports whether the reused conversation was resolved. `ImportChat` calls a new `reopenConversation` only when `MessagesWrote > 0`, after the touch, inside the same transaction. A replay whose every row hits the `source_id` probe reaches neither. The skipped rows still return their link, which `storeChatwootImportLinks` uses to repair the local table -- so the orphaned row heals on the next run without the thread moving.

Covered at both levels: `TestImportChat_ResolvedConversationStaysResolvedWhenEveryRowSkips` (sqlmock, asserts no UPDATE and that the skipped row yields a link) and `TestSyncChatPGIntegration_OrphanedChatwootMessageDoesNotReopen` against the real schema -- imports, deletes the local link, resolves, re-runs `syncChatPG`, and checks the conversation stays resolved, no duplicate row, and the link is back. The inverse guard, `TestImportChat_ReopensResolvedConversationOnlyAfterAWrite`, pins the order: INSERT, touch, then the reopen UPDATE.

**REST path, same defect.** The reopen lived inside `client.FindOrCreateConversation` (`client.go:597`), so `syncChat` and the media pre-pass reopened at resolve time, before any post. That has the same link-durability window (`CreateMessage` succeeds, the link store runs after), and worse: any message that can never land -- expired media, a 4xx the retry policy gives up on -- reopened a resolved thread on every restart until it aged out of the window. `findOrCreateHistoryConversation` now resolves without reopening (the client's lookup minus the toggle), `syncMessageWithOptions` reports whether it posted, and `syncChat` and the pre-pass call `reopenHistoryConversation` once, after the first confirmed post. The live inbound path in `webhook_forward.go` keeps the composite client call, where a message is guaranteed to follow.

`TestSyncChatReopensOnlyAfterFirstPostedMessage` records the request sequence and asserts `toggle_status` comes after the first `POST /messages` and happens exactly once; `TestSyncChatDoesNotReopenWhenEveryPostFails` and `TestRESTMediaPrePassDoesNotReopenWhenNoAttachmentPosts` assert no toggle when nothing lands. With the reopen moved back to resolve time, the first fails with `toggle_status (index 2) came before the first message POST (index 3)`.

Also in this commit: `chatwootLinkFor` is now the single definition of "already linked" for the prefilter and the per-message guard (the two had already drifted); `pendingHistoryMessages` replaces the block duplicated between `syncChat` and `syncChatPG`, and on a lookup failure counts the unclassified messages as failed so `Total > Synced + Failed` no longer happens for an aborted chat; the reopen UPDATE is pinned with a `reopenConversationSQL` constant per `pgimport/AGENTS.md`.

Two things deliberately left as they are:

- A chat whose messages are all linked no longer calls `FindOrCreateContact`, so a group renamed during a quiet window keeps its old Chatwoot name until the next live inbound message refreshes it. Skipping that REST call per chat per restart is the point of the prefilter; flagging it so it is a choice, not an oversight.
- The REST path still has no `source_id` dedupe. In the crash window above it therefore posts the message a second time on the next sync -- and then reopens, correctly, because it wrote. That happens once (the second post stores the link) and it is the duplicate, not the reopen, that is the defect. It predates this PR and needs its own probe in the REST client; I would rather do that as a separate change than fold it in here.

`go vet ./...`, `go test ./...` (13/13), `-race`, the pgimport suite and the four gated real-Postgres tests are green. Still no PR-triggered Actions run for this head, as you noted.

src/infrastructure/chatwoot/sync.go
```diff
@@ -332,25 +303,53 @@ func (s *SyncService) syncChat(
...
 		if err != nil {
 			logrus.Warnf("Chatwoot Sync: Failed to sync message %s: %v", msg.ID, err)
 			progress.IncrementFailedMessages()
 			// Continue with other messages
 		} else {
 			progress.IncrementSyncedMessages()
 		}
+		if posted && !reopened {
+			s.reopenHistoryConversation(conversation)
+			reopened = true
```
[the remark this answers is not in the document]
aguinaldotupy: Fixed in `b95a6c6`. `reopenHistoryConversation` now takes a context, runs the toggle through `retrySyncOp`, and returns whether the thread is in the wanted state; both `syncChat` and the media pre-pass assign that to `reopened` and keep retrying on each later successful post. The pre-pass got the same treatment rather than a single attempt after all posts, since its rows are skipped by pgimport afterwards and nothing downstream would retry. When a pass still ends with messages posted into a thread it could not reopen, it logs at error level -- the one residual, since a REST post is durable before the toggle runs.

`TestSyncChatRetriesReopenOnLaterPostsAfterToggleFailure` and `TestRESTMediaPrePassRetriesReopenOnLaterPostsAfterToggleFailure` fail the toggle three times and assert the second post repairs it on the fourth call. Reverting either loop to the previous unconditional `reopened = true` makes both fail with `toggle_status called 3 times, want 4`. `coderabbit review --uncommitted` reported no findings on the final tree.

src/infrastructure/chatwoot/sync.go
```diff
@@ -332,32 +308,72 @@ func (s *SyncService) syncChat(
...
+			reopened = s.reopenHistoryConversation(ctx, conversation)
+		}
 
 		// Rate limiting: pause between batches
 		if i > 0 && i%opts.BatchSize == 0 {
 			time.Sleep(opts.DelayBetweenBatches)
 		}
 	}
 
+	if anyPosted && !reopened {
```
aldinokemal, line 371: **P1 — a failed reopen is still lost across runs.** By the time this branch is reached, `syncMessageWithOptions` has already created the Chatwoot message and normally persisted its local link. If all `toggle_status` retries fail and this is the last (or only) successful post, the next history sync filters the linked row out before resolving the conversation, so the reopen is never retried and the new message can remain permanently inside a resolved thread. Logging is not a recovery mechanism here. Persist a retryable reopen intent/state (without reposting the message), and cover `POST succeeds -> all toggles fail -> next sync retries toggle with zero duplicate POSTs`. The REST media pre-pass has the same durability gap because pgimport subsequently skips its REST-created row.

aldinokemal (changes requested):
The previous pgimport durability blocker is fixed: reopening now happens only after `MessagesWrote > 0` inside the same transaction, skipped `source_id` rows return links to repair local state, and the real-Postgres regression test covers the missing-local-link crash window. I resolved that thread.

There is still one blocking durability gap in the REST history path. A REST message is created and its local Chatwoot link is stored before `reopenHistoryConversation` runs. If every `toggle_status` retry fails and there is no later successfully posted message in the same pass, this code only logs the failure and returns success. On the next sync the persisted link causes `pendingHistoryMessages` to filter that message out, so nothing retries the reopen; the new customer message can remain permanently inside a resolved conversation. The same failure mode exists in the REST media pre-pass, whose rows are then skipped by pgimport.

Please make the reopen intent durable across runs (for example, persist a retryable reopen state/job keyed by the conversation, or otherwise ensure a linked REST post that failed to reopen is picked up on a later sync without reposting the message). Add a regression test with one successfully-created message, all toggle attempts failing, then a second sync: it should retry the status transition while not creating a duplicate message. The media pre-pass should have equivalent coverage.

No PR-triggered GitHub Actions run is available for this exact head; the exposed commit status is CodeRabbit success.

PR #824, opened 2026-09-03
fix(chat): preserve peer name for outgoing messages

hamedyosefian:
## Context

After pairing account Alice (`+1 555-010-1000`), a chat with saved contact Bob (`+1 555-010-2000`) is initially displayed as Bob. If Alice then sends Bob a message from the primary phone, the linked-device event received by GOWA contains:

- `IsFromMe = true`
- `Chat = 15550102000@s.whatsapp.net`
- `Sender = 15550101000@s.whatsapp.net`
- `PushName = Alice Smith`

`CreateMessage` previously passed the sender's push name to the chat-name resolver. When the existing peer name was a numeric/JID placeholder, it was replaced with `Alice Smith`, so `/chats` and gowa-ui mislabeled Bob's conversation as Alice.

## Fix

For `IsFromMe` events:

- do not use the sender's `PushName` as the peer chat name;
- use the normalized chat JID user as the deterministic fallback.

Incoming-message behavior, group/newsletter naming, LID normalization, and device scoping remain unchanged.

## Test results

- Added `TestCreateMessageFromMeKeepsPeerChatName`.
- Verified the regression test fails before the fix (`"Alice Smith"` replaces Bob's JID placeholder) and passes after it.
- `go test ./... -count=1`
- `go vet ./...`

### DIFF
--- a/src/infrastructure/chatstorage/sqlite_repository.go
+++ b/src/infrastructure/chatstorage/sqlite_repository.go
@@ -1903,8 +1903,17 @@ func (r *SQLiteRepository) CreateMessage(ctx context.Context, evt *events.Messag
 		return r.DeleteMessageByDevice(deviceID, revokedMessageID, chatJID)
 	}
 
+	// PushName belongs to the sender. For messages sent from this account, using
+	// it as the peer chat name would label the recipient with our own name.
+	chatNameSenderUser := normalizedSender.User
+	chatNamePushName := evt.Info.PushName
+	if evt.Info.IsFromMe {
+		chatNameSenderUser = normalizedChatJID.User
+		chatNamePushName = ""
+	}
+
 	// Get appropriate chat name using pushname if available (device-scoped)
-	chatName := r.GetChatNameWithPushNameByDevice(deviceID, normalizedChatJID, chatJID, normalizedSender.User, evt.Info.PushName)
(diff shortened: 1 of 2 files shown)


aldinokemal (approved):
Reviewed current head `025628a09a60861746d6d8f6b1c9673d721b5a09` against current `main` and the pinned `go.mau.fi/whatsmeow@0fadda796019` JID/message-source semantics.

The fix is correctly scoped at the storage naming boundary: for `IsFromMe` events, `MessageSource.Chat` is the destination chat while `MessageSource.Sender`/`PushName` describe this account, so suppressing the sender push name and using the normalized peer chat user as the fallback prevents an outgoing companion event from relabeling the recipient as the account owner. Incoming messages keep the existing push-name upgrade behavior; group/newsletter/status naming remains on the existing helper branches; LID normalization and device-scoped lookup happen before this change; and no storage/API/protocol contract is altered.

The regression test reproduces the important existing-chat case where the stored peer label is still a numeric/JID fallback and verifies an outgoing event cannot replace it with the owner's push name. I did not find a blocking correctness, protocol, device-scoping, or compatibility issue.

No pull-request-triggered GitHub Actions run is attached to this exact head; the available commit status is CodeRabbit success, while the PR reports local `go test ./... -count=1` and `go vet ./...` passing.

PR #826, opened 2026-09-05
fix(mcp): include JSON payload in read tool text results

### LINKED ISSUE #821: MCP read tools return data only in structuredContent — invisible to Claude and other text-only MCP clients
### Describe the bug

MCP read tools (e.g. `whatsapp_chat` → `list_chats`, `list_contacts`, `get_messages`) return their payload only in `structuredContent`. The `content` text block contains just a human-readable summary such as `Retrieved 3 chats (offset 0, limit 3)`.

Clients that do not process `structuredContent` — notably Claude (claude.ai custom connectors / Claude Code) — only show the model the `content` blocks. The model therefore sees "Retrieved 25 chats" but never the chats themselves, and read tools are effectively unusable from Claude. Write tools (`send`, `react`, …) are unaffected because their confirmation string *is* the result, which is why this is easy to miss in normal testing.

The MCP spec recommends serializing the structured payload into a text block precisely for such clients ([Tools — structured content](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content)):

> For backwards compatibility, a tool that returns structured content SHOULD also return the serialized JSON in a TextContent block.

### Reproduction

Against a v9.3.0 instance with `MCP_ENABLED=true` (streamable HTTP, after `initialize`):

```bash
curl -s -u "$APP_BASIC_AUTH" "$BASE/mcp" \
  -H "Content-Type: application/json" \
  -H "Accept: application/json, text/event-stream" \
  -H "Mcp-Session-Id: $SID" \
  -d '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"whatsapp_chat","arguments":{"action":"list_chats","limit":3}}}'
```

Response (trimmed):

```json
{
  "result": {
    "content": [
      { "type": "text", "text": "Retrieved 3 chats (offset 0, limit 3)" }
    ],
    "structuredContent": { "data": [ { "jid": "…", "name": "…" } ], "pagination": { "total": 313 } }
  }
}
```

Connecting the same server to claude.ai as a custom connector and asking for the chat list yields only the summary line — the data in `structuredContent` never reaches the model.

### Affected handlers

All read-style actions in `src/ui/mcp/` that call `mcpg.NewToolResultStructured(resp, fmt.Sprintf(...))` with a summary-only fallback:

- `chat.go`: `list_chats`, `list_contacts`, `get_messages`
- `group.go`: `info`, `participants`, `join_requests`

(Write-style actions also use summary fallbacks, but there the summary is the expected result.)

### Suggested fix

For the read handlers, put the serialized JSON into the text fallback, e.g. use `mcpg.NewToolResultStructuredOnly(resp)` (mcp-go marshals the payload into the text block automatically), or keep the summary and append the JSON. Either variant keeps `structuredContent` for spec-aware clients and makes the data visible to text-only clients.

### Environment

- Version: v9.3.0 (`aldinokemal2104/go-whatsapp-web-multidevice:latest`, digest from 2026-09-02)
- mcp-go v0.56.0
- Client: claude.ai custom connector (OAuth flow itself works fine)

Kizuno18:
## Context

Text-only MCP clients (Claude among them) never read `structuredContent`, and our
read tools put the payload nowhere else — the `content` text block was just a
summary like "Retrieved 3 chats (offset 0, limit 25)". So those tools are
effectively blind on such clients. Closes #821.

Added `structuredWithJSON` in `src/ui/mcp/result.go`, which keeps the existing
human summary and appends the marshalled payload to the same text block (the
structured content is unchanged, so clients that do read it see exactly what they
saw before). If marshalling fails it falls back to the summary alone.

Swapped at the read call sites only:

- `src/ui/mcp/chat.go` — `list_chats`, `list_contacts`, `get_messages`
- `src/ui/mcp/group.go` — `info`, `participants`, `join_requests`
- `src/ui/mcp/app.go` — `status`

Write actions (send, create, join, leave, archive, mark read, participant
changes, ...) still return a summary; their text already describes the outcome
and dumping the response JSON there adds nothing. `invite_link` and
`download_media` are left alone too — the link and the file path are already in
their summary text.

## Test Results

- `gofmt -l .` — clean for `ui/mcp` (the three files it lists,
  `pkg/error/app_error.go`, `usecase/device_test.go`,
  `infrastructure/whatsapp/chatwoot_content_test.go`, are unformatted on main
  already and untouched here)
- `go build ./...`, `go vet ./...`, `go test ./...` all pass from `src/`
- New `src/ui/mcp/result_test.go`: table test over `structuredWithJSON` (map,
  slice, nil and an unmarshalable payload) plus handler tests asserting
  `list_chats` and group `participants` return text containing the serialized
  payload, and that `archive` still returns the summary only

---

### DIFF
--- a/src/ui/mcp/result.go
+++ b/src/ui/mcp/result.go
@@ -1,0 +1,18 @@
+package mcp
+
+import (
+	"encoding/json"
+
+	mcpg "github.com/mark3labs/mcp-go/mcp"
+)
+
+// structuredWithJSON keeps the human summary but appends the JSON payload to
+// the text block: text-only MCP clients never read structuredContent, so a
+// read tool that only summarizes returns nothing usable to them.
+func structuredWithJSON(structured any, summary string) *mcpg.CallToolResult {
+	payload, err := json.Marshal(structured)
+	if err != nil {
+		return mcpg.NewToolResultStructured(structured, summary)
+	}
+	return mcpg.NewToolResultStructured(structured, summary+"\n"+string(payload))
+}
--- a/src/ui/mcp/result.go
+++ b/src/ui/mcp/result.go
@@ -12,7 +12,7 @@ import (
 func structuredWithJSON(structured any, summary string) *mcpg.CallToolResult {
 	payload, err := json.Marshal(structured)
 	if err != nil {
-		return mcpg.NewToolResultStructured(structured, summary)
+		return mcpg.NewToolResultText(summary)
 	}
 	return mcpg.NewToolResultStructured(structured, summary+"\n"+string(payload))
 }
--- a/src/ui/mcp/app.go
+++ b/src/ui/mcp/app.go
@@ -56,7 +56,7 @@ func (h *AppHandler) handleApp(ctx context.Context, request mcpg.CallToolRequest
 			return mcpg.NewToolResultError(err.Error()), nil
 		}
 		structured := map[string]any{"is_connected": isConnected, "is_logged_in": isLoggedIn, "device_id": deviceID}
-		return mcpg.NewToolResultStructured(structured, fmt.Sprintf("connected=%t logged_in=%t", isConnected, isLoggedIn)), nil
+		return structuredWithJSON(structured, fmt.Sprintf("connected=%t logged_in=%t", isConnected, isLoggedIn)), nil
 	case "login_qr":
 		resp, err := h.appService.Login(ctx, deviceID)
 		if err != nil {
--- a/src/ui/mcp/chat.go
+++ b/src/ui/mcp/chat.go
@@ -60,13 +60,13 @@ func (h *ChatHandler) handleChat(ctx context.Context, request mcpg.CallToolReque
 		if err != nil {
 			return mcpg.NewToolResultError(err.Error()), nil
 		}
-		return mcpg.NewToolResultStructured(resp, fmt.Sprintf("Retrieved %d chats (offset %d, limit %d)", len(resp.Data), req.Offset, req.Limit)), nil
+		return structuredWithJSON(resp, fmt.Sprintf("Retrieved %d chats (offset %d, limit %d)", len(resp.Data), req.Offset, req.Limit)), nil
 	case "list_contacts":
 		resp, err := h.userService.MyListContacts(ctx)
 		if err != nil {
 			return mcpg.NewToolResultError(err.Error()), nil
 		}
-		return mcpg.NewToolResultStructured(resp, fmt.Sprintf("Found %d contacts", len(resp.Data))), nil
+		return structuredWithJSON(resp, fmt.Sprintf("Found %d contacts", len(resp.Data))), nil
 	case "get_messages":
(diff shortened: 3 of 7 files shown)


src/ui/mcp/result.go
```diff
@@ -0,0 +1,18 @@
...
+	mcpg "github.com/mark3labs/mcp-go/mcp"
+)
+
+// structuredWithJSON keeps the human summary but appends the JSON payload to
+// the text block: text-only MCP clients never read structuredContent, so a
+// read tool that only summarizes returns nothing usable to them.
+func structuredWithJSON(structured any, summary string) *mcpg.CallToolResult {
+	payload, err := json.Marshal(structured)
+	if err != nil {
+		return mcpg.NewToolResultStructured(structured, summary)
```
aldinokemal, line 15: [P1] This fallback still leaves the unmarshalable value in `CallToolResult.StructuredContent`. In pinned `mcp-go v0.56.0`, that field is emitted as `structuredContent`, so the eventual MCP response marshal fails and the client never receives `summary`. Return `mcpg.NewToolResultText(summary)` here (or otherwise clear structured content), then update the `chan int` case to assert `StructuredContent == nil` and that `json.Marshal(res)` succeeds. The current test only inspects the in-memory text block, so it misses the wire failure.

aldinokemal (changes requested):
The read-tool parity change is well targeted: it keeps `structuredContent` unchanged, only adds equivalent JSON to the text fallback for read actions, and leaves write-action summaries alone. I checked all seven changed files against current main and the pinned `github.com/mark3labs/mcp-go v0.56.0` contract. There is one correctness issue in the helper's advertised fallback path that should be fixed before merge.

`CallToolResult.StructuredContent` is serialized on the wire. If the preflight `json.Marshal(structured)` fails, returning `NewToolResultStructured(structured, summary)` keeps the same unmarshalable value in `StructuredContent`; the final MCP response marshal fails, so the text-only client does not receive the summary at all. Please return a text-only result on that branch and make the existing unmarshalable test marshal the whole result to prove the fallback is actually deliverable.

The normal production payloads at the current call sites are JSON-safe structs/maps, so I did not find a broader architecture/device/protocol concern. There are no PR-triggered workflow runs for this exact head; the author-reported build/vet/test results are not represented by GitHub Actions.

Kizuno18:
pushed baa4c8f

the marshal-error branch now returns `mcpg.NewToolResultText(summary)` so nothing unmarshalable ever lands in `StructuredContent`. the `chan int` case moved to its own subtest and now asserts `res.StructuredContent == nil` and that `json.Marshal(res)` on the whole result succeeds, which is what actually proves the fallback reaches a text-only client

src/ui/mcp/result.go
Kizuno18: Fixed in baa4c8f. The marshal-error branch now returns a text-only result. The regression checks that StructuredContent is nil and the complete CallToolResult can be JSON-encoded.

Issue #823, opened 2026-09-03
🐞 Chats/messages from Business does not appear or sync.


akashgurnani:
### What happened?

WhatsApp messages sent by a business most probably using WABA/WhatsApp Official API does not sync, those chats appear blank.

### Reproduction Steps

```bash
1.Connect WhatsApp
2.Open chats
3.Search for a business chat normally any chat which uses Official WhatsApp API
4.Chat will appear empty, it works on WhatsApp web though.
```

### Relevant log output

```bash

```

### Screenshots or Additional Context

<img width="957" height="391" alt="Image" src="https://github.com/user-attachments/assets/2df7d918-b712-4e6c-a74c-407512f50b36" />

### Application Version

9.3.0

### Verification

- [x] I have checked for existing closed issues and discussions.

aldinokemal:
Hi @akashgurnani, thanks for reporting this. Could you please provide a clearer hint from the GOWA logs while opening or syncing one affected business chat? The “Relevant log output” section is currently empty.

Please include any lines containing one of these clues, if present:
- `Received message` with the message type (for example, `interactiveMessage`)
- `Skipping message ... no content or media`
- `Undecryptable message` or `Unavailable message`

Also let us know whether this happens during the initial history sync or only for new/live messages, and include the exact GOWA version/commit. Please redact phone numbers and message content if needed. These details will help distinguish a message-extraction/storage issue from a message that cannot be decrypted on the linked device.

PR #827, opened 2026-09-05
fix(message): extract text from business template and interactive messages

Kizuno18:
## Context

Chats from WhatsApp Business / Cloud API senders render blank (#823) because
`ExtractMessageTextFromProto` only knows about `conversation`, `extendedTextMessage`,
media captions and the *response* variants of buttons/lists/templates. Business
senders put their body somewhere else entirely, so extraction returned `""`.

That empty string is not cosmetic: both `sqlite_repository.StoreMessage` and
`history_sync` skip a message when content and media type are both empty, so the
message is never stored at all — hence a chat that exists but shows nothing, while
WhatsApp Web renders it fine. Webhook payloads lost `body` the same way.

Message types now handled, preferring body text, then title, then footer:

- `TemplateMessage` — hydrated template and the `hydratedFourRowTemplate` oneof,
  plus the legacy `fourRowTemplate` (content/title/footer HSM) and
  `interactiveMessageTemplate`
- `HighlyStructuredMessage` — via `hydratedHsm`; a non-hydrated HSM only carries an
  element name and substitution params, so it stays empty on purpose
- `InteractiveMessage` — body text, header title/subtitle, footer
- `InteractiveResponseMessage` — body text
- `ButtonsMessage` — content text, header text, footer text
- `ListMessage` — description, title, footer text
- `ProductMessage` — body, product title/description, catalog title, footer
- `OrderMessage` — message, order title

`BuildEventMessage` gets the same fallback so webhook `body` and stored history stay
in parity. Unknown types still return `""`.

One caveat worth flagging: the issue has no raw payload or log, so this is based on
the proto types the Cloud API actually sends rather than a captured repro.
@akashgurnani, could you confirm your business chats fill in after this? If some are
still blank, the exact `Message` field that arrives would pin it down. Marked
`Refs #823` rather than `Closes` for that reason.

## Test Results

From `src/`:

- `gofmt -l .` — only the three files that were already unformatted on `main`
  (`infrastructure/whatsapp/chatwoot_content_test.go`, `pkg/error/app_error.go`,
  `usecase/device_test.go`); nothing new.
- `go build ./...` — clean
- `go vet ./...` — clean
- `go test ./...` — all packages pass

New table-driven cases in `pkg/utils/whatsapp_test.go` cover hydrated templates
(content text and title fallback), the legacy four-row template, hydrated and
non-hydrated HSM, interactive body and header fallback, interactive response,
buttons content and header text, list description and title, product, order, and an
unknown type that must stay empty. Plus one `BuildEventMessage` case for the webhook
path.

No live business-account test — I don't have a WABA sender to send from.

---


src/pkg/utils/whatsapp.go
```diff
@@ -225,6 +225,12 @@ func ExtractMessageTextFromProto(msg *waE2E.Message) string {
 		return templateButtonReply.GetSelectedDisplayText()
 	}
 
+	// Check for business-originated types (template, interactive, buttons,
+	// list, product and order messages sent by Business/Cloud API senders)
+	if businessText := extractBusinessMessageText(msg); businessText != "" {
```
aldinokemal, line 230: [P1] History-sync can still drop the exact business messages this PR is meant to restore. `history_sync.go` passes the raw `WebMessageInfo.Message` directly to `ExtractMessageTextFromProto`, while this function only inspects top-level fields. At the pinned whatsmeow revision (`28bfe537ea6a`), ephemeral/view-once messages are `FutureProofMessage` wrappers; live `events.Message` creation unwraps them, but history-sync does not. A template/interactive message restored from an ephemeral chat therefore still returns empty here and is skipped as `content == "" && mediaType == ""`. Please unwrap at the extractor boundary (`msg = UnwrapMessage(msg)`) or explicitly unwrap the history-sync path, and add a regression test with a wrapped TemplateMessage/InteractiveMessage so storage and webhook behavior stay in parity.

aldinokemal (changes requested):
Verdict: REQUEST CHANGES

The business-message field mapping itself is compatible with the pinned whatsmeow revision (`28bfe537ea6a`) and the webhook path correctly works on an unwrapped message. There is one blocking storage/history gap: history-sync passes raw proto messages into this extractor, so FutureProof wrappers can still make template/interactive messages disappear. Please fix the wrapper boundary and cover it with a regression test before merge.

Kizuno18:
pushed 9a1acad

went with the extractor-boundary option: `ExtractMessageTextFromProto` now calls `UnwrapMessage` first. that covers history sync but also the other raw-message callers with the same gap (`StoreMessage` / edited-message path in `sqlite_repository.go` and `usecase/newsletter.go`). `UnwrapMessage` is idempotent so the callers that already unwrap (`BuildEventMessage`) are unaffected

regression: `TestExtractMessageTextFromProtoWrappedBusinessMessagesMatchWebhook` — TemplateMessage inside `EphemeralMessage` and InteractiveMessage inside `ViewOnceMessage`, asserting the storage path returns the same text as `BuildEventMessage(...).Text`. same two wrapped cases also added to the existing table test

src/pkg/utils/whatsapp.go
Kizuno18: Fixed in 9a1acad. ExtractMessageTextFromProto now unwraps at the extractor boundary, with ephemeral-template and view-once-interactive tests checking parity with BuildEventMessage.

penard-monkey:
You asked for a captured repro rather than inferred proto types — I have one, and it shows this PR fixes the disappearance but still loses most of the message.

### The repro

An inbound Business template (Meta's own "Order confirmed" sample, sent from the WhatsApp Manager test sender) on a linked device running stock `v9.3.1`. Two things happen upstream of this PR:

```
level=warning msg="Unrecognized message type from <redacted>@lid (id=<redacted>):
populated proto fields=[templateMessage messageContextInfo]"
```

1. The `message` webhook fires but carries **no content fields at all** — payload keys were exactly `[chat_id, chat_lid, from, from_lid, id, is_from_me, sender_display_name, timestamp]`.
2. `GET /chat/{jid}/messages` never lists it, because `CreateMessage`'s `content == "" && mediaType == ""` guard skips the row — exactly the mechanism you describe.

So the diagnosis in the PR body is confirmed against a real payload, not just the proto types. The message arrives as `templateMessage` + `messageContextInfo`, with the body in `hydratedTemplate`.

### What this branch produces

I checked out `pull/827/head` and ran that message through `ExtractMessageTextFromProto`. It now returns non-empty, which is what the storage guard keys on — so the disappearance is fixed. But the text it returns is:

```
Hi John Doe,

Thank you for your purchase! Your order number is 123456.
```

The original message has four parts:

| part | value | extracted? |
|---|---|---|
| `hydratedTitleText` | `Order confirmed` | ❌ |
| `hydratedContentText` | `Hi John Doe, …` | ✅ |
| `hydratedFooterText` | `developers.facebook.com` | ❌ |
| `hydratedButtons[0]` (urlButton) | `Visit order details` + URL | ❌ |

### Why

```go
func extractHydratedTemplateText(tpl *waE2E.TemplateMessage_HydratedFourRowTemplate) string {
	return firstNonEmptyText(tpl.GetHydratedContentText(), tpl.GetHydratedTitleText(), tpl.GetHydratedFooterText())
}
```

`firstNonEmptyText` returns the first of the three rather than composing them, so title and footer are only ever reachable when the body is empty. And `GetHydratedButtons()` isn't read anywhere in the diff, so the CTA is dropped in every case.

For transactional templates — order confirmations, shipping notices, appointment reminders — the title and the CTA are usually the point of the message, so this reads as a partial message rather than a blank one.

### Suggestion

Compose the rows instead of taking the first, and render the buttons. `formatNativeFlowButton` in `infrastructure/whatsapp/webhook_forward.go` already establishes the labelling for the interactive equivalents (`🔗 <text>: <url>`, `📞 <text>: <phone>`), so the template buttons can match it:

```go
func formatHydratedTemplate(tpl *waE2E.TemplateMessage_HydratedFourRowTemplate) string {
	var parts []string
	if title := tpl.GetHydratedTitleText(); title != "" {
		parts = append(parts, title)
	}
	if content := tpl.GetHydratedContentText(); content != "" {
		parts = append(parts, content)
	}
	if footer := tpl.GetHydratedFooterText(); footer != "" {
		parts = append(parts, footer)
	}
	for _, button := range tpl.GetHydratedButtons() {
		if line := formatHydratedTemplateButton(button); line != "" {
			parts = append(parts, line)
		}
	}
	return strings.Join(parts, "\n")
}
```

…with `formatHydratedTemplateButton` switching on `GetUrlButton()` / `GetCallButton()` / `GetQuickReplyButton()`.

One caveat on composing: a template whose parts are all empty would then extract as `""` and be skipped from storage again. It's worth returning a sentinel (`"Template message"`, matching `formatInteractiveMessageSummary`'s `"Interactive message"`) for a non-nil template, so an unreadable template still can't vanish from the thread.

I have this running locally against v9.3.1 with tests covering the four parts and the three button kinds. Happy to open it as a PR against your branch, or leave it here if you'd rather fold it in yourself.

src/pkg/utils/whatsapp.go
```diff
@@ -1142,6 +1274,10 @@ func BuildEventMessage(evt *events.Message) (message EvtMessage) {
 		}
 	}
 
+	if message.Text == "" {
+		message.Text = extractBusinessMessageText(msg)
```
aldinokemal: [P1] This breaks Chatwoot rendering for interactive messages.
- `buildChatwootMessageContent` (`webhook_forward.go:562`) only uses the pre-rendered `payload["interactive"]` summary from #794 when `body` is empty.
- This line now sets `body` to the interactive body text.
- So the header, footer, CTA buttons and carousel cards never reach Chatwoot, on both the live path and the retry-queue path:

| message | main | this PR |
|---|---|---|
| CTA | `Order confirmed / Hi John… / acme.com / 🔗 Track order: https://…` | `Hi John…` |
| carousel | `Pick a plan / Card 1: … 🔗 …` | `Pick a plan` |
| header image caption | `Summer sale / 50% off today` | `50% off today` |

The existing tests miss this because they build payloads by hand with `interactive` and no `body`.

The fix below keeps the webhook body and storage as they are, and has Chatwoot prefer the summary. It restores main's output exactly:

```go
if interactive, ok := data["interactive"].(string); ok && interactive != "" {
	content = interactive
} else if body, ok := data["body"].(string); ok && body != "" {
	content = utils.WhatsAppToChatwootMarkdown(body)
}
```

Please add a test that runs an interactive message with a header, a footer and a `cta_url` button through `buildEventPayload` and then `buildChatwootMessageContent`.

src/pkg/utils/whatsapp.go
```diff
@@ -178,6 +178,10 @@ func ExtractMessageTextFromProto(msg *waE2E.Message) string {
 		return ""
 	}
 
+	// Unwrap FutureProof wrappers (ephemeral, view-once) so callers that pass
+	// raw WebMessageInfo.Message (e.g. history sync) see the inner content.
+	msg = UnwrapMessage(msg)
```
aldinokemal: [P1] Because only the text side is unwrapped, history sync now damages stored media.
- `history_sync.go:256-257` passes the same raw `msg.GetMessage()` to this function and to `ExtractMediaInfo`. `ExtractMediaInfo` still reads top-level fields only.
- So for an ephemeral or view-once image, video or document with a caption, this function now returns the caption while `mediaType` stays `""`.
- On main that message was skipped. Now it gets written. `StoreMessagesBatch` overwrites `media_type`, `url`, `direct_path`, `media_key` and the other media columns without checking them first.
- So when history delivers a message again (for example after re-pairing, which keeps chat history), the media fields of the row the live path stored are cleared. New rows are stored as caption-only text.

Reproduced with a real SQLite repo and `processConversationMessages`:
- After the live store: `media_type="image"`.
- After a history re-sync of the same ID as `EphemeralMessage{ImageMessage}`: `media_type=""`, `url=""`, `direct_path=""`.
- On main the row is left intact.

Fix: unwrap once in `history_sync.go` and pass the result to both extractors:

```go
inner := utils.UnwrapMessage(msg.GetMessage())
content := utils.ExtractMessageTextFromProto(inner)
mediaType, filename, mediaURL, directPath, mediaKey, fileSHA256, fileEncSHA256, fileLength := utils.ExtractMediaInfo(inner)
```

This also means wrapped media from history sync is stored with its media. Please add a history-sync test with an ephemeral image that has a caption.

src/pkg/utils/whatsapp.go
```diff
@@ -252,6 +262,128 @@ func ExtractMessageTextFromProto(msg *waE2E.Message) string {
...
+			extractHighlyStructuredText(fourRow.GetFooter()),
+		); text != "" {
+			return text
+		}
+	}
+
+	return extractInteractiveMessageText(tpl.GetInteractiveMessageTemplate())
+}
+
+func extractHydratedTemplateText(tpl *waE2E.TemplateMessage_HydratedFourRowTemplate) string {
```
aldinokemal: [P2] Taking the first non-empty field means body-only in practice.
- Meta requires a body on every template ("The only required component is the body component", see [template components](https://developers.facebook.com/docs/whatsapp/business-management-api/message-templates/components)).
- So for Cloud API traffic the title and footer fallbacks never run, and `GetHydratedButtons()` is never read.
- penard-monkey's capture above shows the result: the title, the footer and the URL CTA are all dropped.

This string becomes the webhook `body`, the stored content and the Chatwoot text. It's better to settle the format now than to change `body` a second time later.

Prior art that renders these messages for display joins the parts. mautrix-whatsapp's `convertTemplateMessage` ([wa-business.go](https://github.com/mautrix/whatsapp/blob/main/pkg/msgconv/wa-business.go)) builds the title, content, button lines and footer. Only Baileys' `extractMessageContent`, which normalises content rather than rendering it, returns the body alone.

I'd go with penard-monkey's suggestion:
- Join the title, content and footer with `\n`.
- Render `hydratedButtons` in the `formatNativeFlowButton` style (`🔗 <text>: <url>`, `📞 <text>: <phone>`).

The same applies to list and buttons messages, at lower priority.

aldinokemal (changes requested):
Verdict: REQUEST CHANGES

The unwrap fix is in, and the business field mapping matches the pinned whatsmeow. This is the right direction for #823. Two regressions in existing behaviour need fixing before merge. One design point should be settled in this PR, because it defines the webhook `body` format.

**Must fix**
1. [P1] Chatwoot no longer shows the rendered interactive summary (header, footer, CTA buttons, carousel cards). See the inline comment on `BuildEventMessage`.
2. [P1] History sync now wipes media metadata on ephemeral and view-once media that has a caption. See the inline comment on the new `UnwrapMessage` call.

**Settle in this PR**
3. [P2] Templates return only the body. See the inline comment on `extractHydratedTemplateText`.

**Minor (these can be follow-ups)**
- The PR body says "webhook `body` and stored history stay in parity", but that isn't true for `buttonsResponse`, `listResponse` and `templateButtonReply`. These are the replies to the messages this PR adds. Storage extracts their text, but they still produce no webhook `body` (so Chatwoot shows `(Unsupported message type)`). This already happens on main, but it's easy to fix here by moving those three branches into `extractBusinessMessageText`.
- In Chatwoot, list and order messages change from `List: <title>` / `Order: <title>` to just the description or message text.
- A template or interactive message with no extractable text still returns `""` and is skipped. Returning a placeholder such as `"Template message"` (like the existing `"Interactive message"`) would fix that.
- `docs/webhook-payload.md` has no section for business messages, although `body` is now populated for 8 more message types.
- Tests:
  - The two wrapped cases appear in both the table test and the parity test. The parity test can't fail on its own, because both paths call `extractBusinessMessageText`.
  - Some branches have no test cases: `interactiveMessageTemplate`, the four-row template footer, and the product fallbacks.
  - There's no end-to-end test from `buildEventPayload` to `buildChatwootMessageContent`.
  - `firstNonEmptyText` can be replaced with the standard library's `cmp.Or`.
- The comments say gowa's `UnwrapMessage` unwraps live events before they arrive. That's actually whatsmeow's `UnwrapRaw`.
- Messages that were already dropped won't reappear after upgrading. A chat with no stored messages can't use `/chat/{jid}/history` either. So @akashgurnani should test with a new business message or by re-linking the device. Keeping `Refs` rather than `Closes` is correct.

Verified on the PR head (49a2e09):
- `go build` and `go vet` pass, and so do the tests for `pkg/utils`, `infrastructure/whatsapp`, `infrastructure/chatstorage` and `infrastructure/chatwoot/...`.
- The PR merges cleanly onto current main (1e624e7).

Issue #796, opened 2026-08-14
⚡️ Call logs


FireBird22:
### What problem does your feature request solve?

Allows access to call logs remotely, granting access to information previously hard to get.

### Describe the solution you are proposing.

Since whatsapp web recently added call logs, and now they are accessible through the web app session, I thought it would a good suggestion to add the ability to get these logs

### Importance

Nice to Have

### Additional Context

- [ ] This missing feature presents as a bug.
- [ ] This missing feature is forcing me to consider alternatives.

PR #828, opened 2026-09-05
feat(call): add call logs listing endpoint

Kizuno18:
## Context

Closes the read side of #796: incoming calls are already persisted, but there is
no way to get at them without paging every chat.

`handleCallOffer` stores each `call.offer` through `CreateIncomingCallRecord` as
a synthetic message row with `media_type = "call"` and a small `call_metadata`
JSON blob. Today the only way to read those back is `GET /chat/:jid/messages`
per chat, filtering `media_type == "call"` client-side — which means you need to
know the chat first, and you page through unrelated messages to find calls.

`GET /call/logs` lists those records for the device in context, newest first,
with `limit` (default 25, max 100), `offset` and an optional `chat_jid` filter.
`call_metadata` is parsed into fields instead of being handed back as a JSON
string, so `call_id` can be fed straight into `POST /call/reject`.

Layering follows the existing chat list path: handler parses the query,
`ValidateListCallLogs` applies the limit/offset bounds, and the usecase resolves
the device from context before calling storage. Storage side is a new
`GetCallRecords(*CallRecordFilter)` on `IChatStorageRepository`, the SQLite
repository and `chatstorage_wrapper.go`, filtering `device_id = ?` plus
`media_type = 'call'` and returning the unpaginated total for the pagination
echo. No migration — it queries rows that already exist.

```
$ curl -s -H 'X-Device-Id: 628123456789@s.whatsapp.net' \
    'http://localhost:3000/call/logs?limit=2'
{
  "code": "SUCCESS",
  "message": "Success get call logs",
  "results": {
    "data": [
      {
        "id": "call:ABC123DEF456",
        "chat_jid": "6289685028129@s.whatsapp.net",
        "sender_jid": "6289685028129@s.whatsapp.net",
        "timestamp": "2026-08-22T10:05:00Z",
        "is_from_me": false,
        "call_metadata": {
          "call_id": "ABC123DEF456",
          "auto_rejected": true,
          "remote_platform": "android"
        }
      }
    ],
    "pagination": { "limit": 2, "offset": 0, "total": 12 }
  }
}
```

Deliberately not covered, since it needs event handling this PR does not touch:

- outgoing calls — only `events.CallOffer` is wired up, so `is_from_me` is
  always false today
- duration and outcome (answered/ended/missed) — `CallAccept` and
  `CallTerminate` aren't handled, so there is nothing to report
- backfill of calls from before the device was linked; history sync does not
  carry call records
- no MCP tool, and no `search` param (the stored content is a fixed
  "Incoming call" string, so matching on it buys nothing)

This is a first slice: it exposes what is already stored. Enriching
`call_metadata` from the other call events is the natural follow-up, and the
endpoint shape doesn't change when it lands.

## Test Results

From `src/`:

- `gofmt -l .` — no new entries (`pkg/error/app_error.go`,
  `usecase/device_test.go` and `infrastructure/whatsapp/chatwoot_content_test.go`
  are already unformatted on main)
- `go build ./...` — ok
- `go vet ./...` — clean
- `go test ./...` — all packages pass

New tests:

- `src/infrastructure/chatstorage/sqlite_repository_call_test.go` — against a
  real in-memory-ish SQLite repo: only `media_type = 'call'` rows come back,
  only for the requested device, ordered newest first; limit/offset paging and
  the `chat_jid` filter; missing `device_id` is rejected
- `src/validations/call_validation_test.go` — table test for the limit default
  (25), the 1..100 bounds and the offset floor

---

### DIFF
--- a/src/infrastructure/chatstorage/sqlite_repository.go
+++ b/src/infrastructure/chatstorage/sqlite_repository.go
@@ -2348,6 +2348,71 @@ func (r *SQLiteRepository) CreateIncomingCallRecord(ctx context.Context, evt *ev
 	return r.StoreMessage(message)
 }
 
+// GetCallRecords retrieves stored call records (synthetic messages with
+// media_type "call") newest first, together with the unpaginated total.
+func (r *SQLiteRepository) GetCallRecords(filter *domainChatStorage.CallRecordFilter) ([]*domainChatStorage.Message, int64, error) {
+	if filter == nil || filter.DeviceID == "" {
+		return nil, 0, fmt.Errorf("device_id is required for call record queries (data isolation)")
+	}
+
+	conditions := []string{"device_id = ?", "media_type = 'call'"}
+	args := []any{filter.DeviceID}
+
+	if filter.ChatJID != "" {
+		conditions = append(conditions, "chat_jid = ?")
+		args = append(args, filter.ChatJID)
+	}
+
+	whereClause := " WHERE " + strings.Join(conditions, " AND ")
+
+	total, err := r.getCount("SELECT COUNT(*) FROM messages"+whereClause, args...)
+	if err != nil {
+		return nil, 0, err
+	}
+
+	query := `
+		SELECT id, chat_jid, device_id, sender, content, timestamp, is_from_me,
+			media_type, call_metadata, filename, url, direct_path, media_key, file_sha256,
+			file_enc_sha256, file_length, referral_metadata, created_at, updated_at
+		FROM messages` + whereClause + `
+		ORDER BY timestamp DESC
+	`
+
+	if filter.Limit > 0 {
+		if filter.Limit > 1000 {
+			filter.Limit = 1000
+		}
+		query += " LIMIT ?"
+		args = append(args, filter.Limit)
+
+		if filter.Offset > 0 {
+			query += " OFFSET ?"
+			args = append(args, filter.Offset)
+		}
+	}
+
+	rows, err := r.db.Query(query, args...)
+	if err != nil {
+		return nil, 0, err
+	}
+	defer rows.Close()
+
+	var messages []*domainChatStorage.Message
+	for rows.Next() {
+		message, err := r.scanMessage(rows)
+		if err != nil {
+			return nil, 0, err
+		}
+		messages = append(messages, message)
+	}
+
+	if err := rows.Err(); err != nil {
+		return nil, 0, err
+	}
+
+	return messages, total, nil
+}
+
 // GetStorageStatistics returns current storage statistics for logging purposes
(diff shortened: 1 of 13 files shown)


src/infrastructure/chatstorage/sqlite_repository.go
```diff
@@ -2348,6 +2348,71 @@ func (r *SQLiteRepository) CreateIncomingCallRecord(ctx context.Context, evt *ev
...
+	if err != nil {
+		return nil, 0, err
+	}
+
+	query := `
+		SELECT id, chat_jid, device_id, sender, content, timestamp, is_from_me,
+			media_type, call_metadata, filename, url, direct_path, media_key, file_sha256,
+			file_enc_sha256, file_length, referral_metadata, created_at, updated_at
+		FROM messages` + whereClause + `
+		ORDER BY timestamp DESC
```
aldinokemal, line 2378: [P1] `ORDER BY timestamp DESC` is not a stable pagination order. Multiple call records can have the same timestamp, while each synthetic row has an independent immutable `id`; with `limit`/`offset`, SQLite is free to reorder ties and clients can see duplicated or skipped calls across page requests. Make this a total deterministic order (for example `ORDER BY timestamp DESC, id DESC`) and add a pagination regression test with equal timestamps spanning page boundaries.

aldinokemal (changes requested):
Verdict: REQUEST CHANGES

The endpoint is correctly device-scoped through the request context and repository filter, the repository/wrapper/interface changes are synchronized, and the stored call metadata shape matches `CreateIncomingCallRecord`. One blocking pagination correctness issue remains: the advertised offset pagination does not define a total order for equal timestamps.

Kizuno18:
pushed 391a7fd + 66acfdb

- `GetCallRecords` now orders by `timestamp DESC, id DESC`. no postgres variant of this query exists, only `SQLiteRepository`. regression: `TestGetCallRecordsPaginatesDeterministicallyWithEqualTimestamps` — 5 rows with the same timestamp, paged limit=3 twice, asserts no dup/skip and the exact id order across the boundary
- also took the coderabbit points on `call.go`: `c.Bind().Query(&request)` instead of `fiber.Query[int]` so a malformed `limit`/`offset` is a 400 instead of silently becoming the default (default 25 still applied in `ValidateListCallLogs`), and `Cache-Control: no-store` on the response since it's device-scoped

src/infrastructure/chatstorage/sqlite_repository.go
Kizuno18: Fixed in 391a7fd and 66acfdb. Call pagination now uses timestamp DESC, id DESC. The equal-timestamp regression checks the exact order across two pages and rejects duplicates.

PR #829, opened 2026-09-05
feat: expose referral metadata in chat messages

Kizuno18:
## Context

Rebase of #641 by @juliomuhlbauer onto current main, with the test the review
asked for. His commit is cherry-picked, so authorship stays with him.

`referral_metadata` (Meta Ads / Click-to-WhatsApp attribution) has been stored
since migration 16 and both storage read paths — `GetMessages` and
`SearchMessages` — already select it, but `GET /chat/:jid/messages` never
returned it. This adds `ReferralMetadata` to `domainChat.MessageInfo`, maps it
in `GetChatMessages`, and documents it in the `ChatMessage` OpenAPI schema.

What changed against the original PR:

- Conflict resolution in the `MessageInfo` literal keeps everything main has
  gained since April — `SenderDisplayName` and the `Reactions` block — and only
  adds the one `ReferralMetadata: message.ReferralMetadata` line, so nothing in
  the response regresses.
- The openapi hunk moved with the schema: the original patched around line 4233,
  the `ChatMessage` schema now lives at ~5374. The property sits next to
  `call_metadata`, matching the Go field order.
- gofmt'd the struct — the original left `CallMetadata`/`ReferralMetadata`
  misaligned, which was CodeRabbit's nit.
- Parity check: `MessageInfo` is still built in exactly one place. The MCP
  `whatsapp_chat` / `messages` action returns the same
  `GetChatMessagesResponse` as structured output, so it picks the field up for
  free; the websocket path doesn't emit `MessageInfo`. No other wiring needed.

The field is typed `string` with `omitempty`, same as the sibling
`CallMetadata` — the raw JSON blob is passed through rather than parsed.

Supersedes #641.

## Test Results

`src/usecase/chat_test.go` gets `TestGetChatMessagesExposesReferralMetadata`: it
runs `GetChatMessages` over a stub repo holding one message with referral
metadata and one without, asserts the mapping, then marshals the response and
checks `referral_metadata` is present on the first and absent on the second, so
the `omitempty` tag is covered too. Dropping the mapping line makes it fail.

From `src/`:

- `gofmt -l .` — no new entries (three files are already unformatted on main and
  are left alone: `infrastructure/whatsapp/chatwoot_content_test.go`,
  `pkg/error/app_error.go`, `usecase/device_test.go`)
- `go build ./...` — pass
- `go vet ./...` — pass
- `go test ./...` — pass, 13 packages ok

---

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -5374,6 +5374,10 @@ components:
           type: string
           example: '{"call_id":"ABC","auto_rejected":false}'
           description: JSON string with call details when media_type is `call`. Omitted when not applicable (matches Go json omitempty).
+        referral_metadata:
+          type: string
+          example: '{"ctwa_clid":"clid_123","source_app":"whatsapp"}'
+          description: JSON string with Meta Ads referral/attribution details when the message originated from a Click-to-WhatsApp ad. Omitted when not applicable (matches Go json omitempty).
         filename:
           type: string
           example: 'photo.jpg'
--- a/src/domains/chat/chat.go
+++ b/src/domains/chat/chat.go
@@ -66,12 +66,13 @@ type MessageInfo struct {
 	MediaType         string         `json:"media_type"`
 	Reactions         []ReactionInfo `json:"reactions,omitempty"`
 	// CallMetadata is JSON when media_type is "call" (incoming call log).
-	CallMetadata string `json:"call_metadata,omitempty"`
-	Filename     string `json:"filename"`
-	URL          string `json:"url"`
-	FileLength   uint64 `json:"file_length"`
-	CreatedAt    string `json:"created_at"`
-	UpdatedAt    string `json:"updated_at"`
+	CallMetadata     string `json:"call_metadata,omitempty"`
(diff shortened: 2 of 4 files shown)


aldinokemal (approved):
Reviewed against current main (`3b89617061b6bbd94a7d5c6799c19d923e7f71a0`). This is the clean current-main rebase of the older referral-metadata PR: storage already persists and reads `Message.ReferralMetadata`, and this change only carries that existing device-scoped value through `GetChatMessages` into the REST/MCP response contract.

The response initializer preserves current-main `SenderDisplayName` and reactions, the new field is additive with `omitempty`, and the OpenAPI schema matches the Go representation (raw referral JSON remains a string, like `call_metadata`). The regression test covers both populated and omitted JSON behavior. No migrations, whatsmeow/protobuf behavior, session lifecycle, webhook, websocket, or storage partitioning semantics change here.

I found no blocking correctness, compatibility, or security issue. There are no PR-triggered GitHub Actions runs for this exact head, so the author-reported build/vet/test results are not independently represented by Actions.

PR #831, opened 2026-09-05
feat(webhook): ignore group messages per instance, not just globally

Kizuno18:
Picks up #756 by @huboperacional: rebased onto main and adds the null-reset the review
asked for. His eight commits are cherry-picked as-is, so authorship stays with him; my
three follow-ups sit on top.

## Context

The feature itself is unchanged from #756: a nullable per-device `webhook_ignore_groups`
override, stored on the `devices` table, surfaced on `PATCH`/`GET
/devices/{device_id}/webhook`, and consulted by the webhook forwarder only for `@g.us`
JIDs. `NULL` means "never configured" and the device keeps following the global
`WHATSAPP_WEBHOOK_IGNORE_JIDS` list.

What changed relative to the original PR:

- **Migration renumbered 44 -> 45.** Main grew its own migration 44 (poll definitions)
  while #756 was open, so the `ALTER TABLE devices ADD COLUMN webhook_ignore_groups` is
  now appended as 45. The only other rebase conflict was an append clash in
  `sqlite_repository_test.go` / `device_test.go`; both sides were kept.

- **`null` now clears the override.** This was the maintainer's outstanding item:
  binding the field into a `*bool` collapses "key absent" and "key present but `null`",
  so once a device had `true` or `false` there was no request that could put the column
  back to `NULL`. The handler decodes into `utils.Nullable[bool]` (`Set`/`Valid`/`Value`,
  a small `json.Unmarshaler` in `src/pkg/utils/nullable.go`) and the three states now
  mean what you would expect:

  ```json
  {"webhook_url": "https://hook.example.com"}                                 // keeps the stored override
  {"webhook_url": "https://hook.example.com", "webhook_ignore_groups": null}  // clears it -> global list applies
  {"webhook_url": "https://hook.example.com", "webhook_ignore_groups": false} // this device keeps its groups
  ```

  The repository already wrote a nil `*bool` as SQL `NULL`, so nothing changed below the
  handler.

- **Codex's multi-slot lookup ambiguity is real, and fixed.** The webhook payload's
  `device_id` is the bare NonAD JID, and `GetDeviceRecordByJID` deliberately returns nil
  once two slots share that number (#760) — so on exactly the multi-instance setup this
  feature is for, sibling slots resolved no override at all. The forwarder now resolves
  the record by the AD JID of the slot in the event context (`DeviceFromContext`, set per
  event in `event_handler.go`), which addresses one row, and falls back to the bare JID
  for slots with no AD JID recorded yet (logging the lookup error, not swallowing it).
  Same pass also resolves the device record once instead of twice, so the webhook config
  and the group override can no longer come from two different rows; that left
  `getWebhookConfigForDevice` without a production caller, so it is gone and its tests
  now exercise the live path.

- **An opt-out no longer un-ignores exact group JIDs.** In #756 `webhook_ignore_groups:
  false` skipped the ignore-list match for group JIDs altogether, so a group the operator
  had listed by its exact JID in `WHATSAPP_WEBHOOK_IGNORE_JIDS` came back. The override
  now replaces only the `@g.us` wildcard, which is the contract the review asked for and
  the one the docs describe (`utils.MatchesExactIgnoredJID`).

- **Docs.** `webhook_ignore_groups` was missing from `docs/openapi.yaml` entirely; it is
  now on the PATCH body and on both response bodies, with the tri-state spelled out.
  `docs/webhook-payload.md` gets a bullet in the ignore-JID behaviour list. `readme.md`
  documents only the global env/flag, so it is untouched.

Supersedes #756.

## Test Results

From `src/`:

- `gofmt -l .` — only the three files already unformatted on main
  (`pkg/error/app_error.go`, `usecase/device_test.go`,
  `infrastructure/whatsapp/chatwoot_content_test.go`).
- `go build ./...` — ok
- `go vet ./...` — ok
- `go test ./...` — all packages pass, including huboperacional's original REST, storage
  and forwarder tests unchanged.

New tests on top of his:

- `TestUpdateDeviceWebhook_IgnoreGroupsTriState` — omitted key keeps the stored value,
  `null` clears it, explicit `false` sets it (the `true -> null -> nil` cycle the review
  asked for).
- `TestSQLiteRepositoryDeviceWebhookConfig_IgnoreGroupsClearsToNull` — a set override
  written back as nil round-trips as `NULL` through both `GetDeviceWebhookConfig` and
  `GetDeviceRecordByJID`.
- `TestWebhookIgnoreJID_DeviceOverrideResolvedByADJID` — with the bare number ambiguous,
  the override is still found via the emitting slot's AD JID.
- `TestWebhookIgnoreJID_DeviceOverrideFalseScopeIsTheWildcard` — with `false`, an exact
  group JID in the global list is still ignored while the `@g.us` wildcard is neutralized.
- `TestNullableUnmarshalJSON` / `TestNullableMarshalJSON` / `TestNullableRejectsMismatchedType`,
  `TestMatchesExactIgnoredJID`.

---

### DIFF
--- a/src/infrastructure/whatsapp/webhook_forward.go
+++ b/src/infrastructure/whatsapp/webhook_forward.go
@@ -99,16 +99,17 @@ func getContactMutex(phone string) *sync.Mutex {
 // successful targets still receive the event.
 func forwardPayloadToConfiguredWebhooks(ctx context.Context, payload map[string]any, eventName string) error {
 	deviceJID, _ := payload["device_id"].(string)
-	webhookConfig, err := getWebhookConfigForDevice(deviceJID)
+	record, err := resolveWebhookDeviceRecord(ctx, payload)
 	if err != nil {
 		// A config lookup failure is not a delivery failure: fall back to the global
 		// webhook config so the event still reaches the global targets and Chatwoot.
 		logrus.Warnf("Failed to get webhook config for device %s, falling back to global config: %v", deviceJID, err)
-		webhookConfig = nil
+		record = nil
 	}
+	webhookConfig := webhookConfigFromRecord(record)
 
 	webhookAllowed := isEventWhitelistedForDevice(eventName, webhookConfig) &&
-		!shouldIgnoreWebhookJID(payload)
+		!shouldIgnoreWebhookJID(payload, deviceIgnoreGroupsOverride(record))
 	chatwootAllowed := config.ChatwootEnabled && shouldForwardEventToChatwoot(eventName) && isEventWhitelistedForChatwoot(eventName)
 
 	if !webhookAllowed && !chatwootAllowed {
@@ -171,18 +172,50 @@ func getWebhookConfigForDevice(deviceJID string) (*domainChatStorage.DeviceWebho
 	if err != nil {
 		return nil, fmt.Errorf("failed to get device record: %w", err)
 	}
-	if record != nil && record.WebhookURL != nil && *record.WebhookURL != "" {
-		logrus.Debugf("Using device-specific webhook config for %s", deviceJID)
-		return &domainChatStorage.DeviceWebhookConfig{
-			WebhookURL:                record.WebhookURL,
-			WebhookSecret:             record.WebhookSecret,
-			WebhookEvents:             record.WebhookEvents,
-			WebhookInsecureSkipVerify: record.WebhookInsecureSkipVerify,
-			WebhookIgnoreGroups:       record.WebhookIgnoreGroups,
-		}, nil
+	return webhookConfigFromRecord(record), nil
+}
+
+// webhookConfigFromRecord maps a device registration onto its webhook configuration,
+// returning nil when the device has no device-specific webhook URL so the caller keeps
+// using the global config.
+func webhookConfigFromRecord(record *domainChatStorage.DeviceRecord) *domainChatStorage.DeviceWebhookConfig {
+	if record == nil || record.WebhookURL == nil || *record.WebhookURL == "" {
+		return nil
+	}
+	logrus.Debugf("Using device-specific webhook config for %s", record.DeviceID)
+	return &domainChatStorage.DeviceWebhookConfig{
+		WebhookURL:                record.WebhookURL,
+		WebhookSecret:             record.WebhookSecret,
+		WebhookEvents:             record.WebhookEvents,
+		WebhookInsecureSkipVerify: record.WebhookInsecureSkipVerify,
+		WebhookIgnoreGroups:       record.WebhookIgnoreGroups,
+	}
+}
+
+// resolveWebhookDeviceRecord resolves the registration of the slot that emitted the event.
+// The payload's device_id is the bare NonAD JID, which GetDeviceRecordByJID deliberately
+// refuses to resolve once several slots share one number (issue #760); the AD JID of the
+// slot carried in the event context addresses exactly one row, so try that first and only
+// fall back to the bare JID when the slot has no AD JID recorded yet.
+func resolveWebhookDeviceRecord(ctx context.Context, payload map[string]any) (*domainChatStorage.DeviceRecord, error) {
+	deviceJID, _ := payload["device_id"].(string)
+
+	if inst, ok := DeviceFromContext(ctx); ok && inst != nil {
+		if adJID := inst.ADJID(); adJID != "" && adJID != deviceJID {
+			if record, err := getDeviceRecordForTest(adJID); err == nil && record != nil {
+				return record, nil
+			}
+		}
 	}
 
-	return nil, nil
+	if deviceJID == "" {
+		return nil, nil
+	}
+	record, err := getDeviceRecordForTest(deviceJID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get device record: %w", err)
+	}
+	return record, nil
 }
 
 // getWebhookURLsFromConfig extracts webhook URLs from the config.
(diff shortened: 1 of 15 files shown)


src/infrastructure/whatsapp/webhook_forward.go
```diff
@@ -99,16 +99,17 @@ func getContactMutex(phone string) *sync.Mutex {
...
 func forwardPayloadToConfiguredWebhooks(ctx context.Context, payload map[string]any, eventName string) error {
 	deviceJID, _ := payload["device_id"].(string)
-	webhookConfig, err := getWebhookConfigForDevice(deviceJID)
+	record, err := resolveWebhookDeviceRecord(ctx, payload)
 	if err != nil {
 		// A config lookup failure is not a delivery failure: fall back to the global
 		// webhook config so the event still reaches the global targets and Chatwoot.
 		logrus.Warnf("Failed to get webhook config for device %s, falling back to global config: %v", deviceJID, err)
-		webhookConfig = nil
+		record = nil
```
aldinokemal, line 107: [P1] Fail closed for the generic webhook when this event is a group and device-record resolution fails. Setting `record = nil` here removes an explicit `WebhookIgnoreGroups=true`; if the global ignore list does not contain `@g.us`, the group payload is then forwarded even though this device was configured to suppress it. Keep Chatwoot/non-group fallback behavior if desired, but do not send a generic group webhook until the originating slot's policy is known. Please add a regression test with per-device `true`, no global `@g.us`, and a resolver/storage error, asserting no generic webhook delivery.

src/ui/rest/device.go
```diff
@@ -229,14 +232,23 @@ func (handler *Device) UpdateDeviceWebhook(c fiber.Ctx) error {
...
 	config := &chatstorage.DeviceWebhookConfig{
 		WebhookURL:                req.WebhookURL,
 		WebhookSecret:             req.WebhookSecret,
 		WebhookEvents:             req.WebhookEvents,
 		WebhookInsecureSkipVerify: req.WebhookInsecureSkipVerify,
+		WebhookIgnoreGroups:       ignoreGroups,
 	}
 
-	err := handler.Service.SetDeviceWebhookConfig(c.Context(), deviceID, config)
+	err = handler.Service.SetDeviceWebhookConfig(c.Context(), deviceID, config)
```
aldinokemal, line 251: [P1] The promised “omitted means preserve the stored value” semantics are not atomic here. Request A can read `true`, request B can set `false`, then A can write the stale `true` back as part of this full-row update. Move the tri-state preservation into a single repository/service update (e.g. a `fieldSet` flag/SQL expression, transaction with concurrency control, or revision check) so omission preserves the value that exists at write time. Add an interleaving/concurrency regression test.

aldinokemal (changes requested):
Two blocking issues remain before this is safe to merge. The AD-JID-first lookup direction is correct for multi-slot routing and matches the pinned whatsmeow/current DeviceInstance identity model, and migration 45 is append-only on current main. However, the generic webhook path currently fails open on a device-record lookup error, which can bypass an explicit per-device group suppression policy, and the tri-state PATCH preservation is a non-atomic read/modify/write that can overwrite a concurrent explicit update. Please fix both and add regression coverage for the failure/interleaving cases.

Kizuno18:
pushed 6e132b6 + 21bfc63

- **fail closed**: `forwardPayloadToConfiguredWebhooks` now blocks the generic webhook for group events whenever the device record couldn't be resolved (new `payloadIsGroupEvent`, same chat_id/from/chat_lid/from_lid keys as `shouldIgnoreWebhookJID`). non-group events keep the global fallback, chatwoot path untouched. tests: `TestWebhookIgnoreJID_ResolverErrorFailsClosedForGroupEvent` (per-device true, no global `@g.us`, storage error → no delivery) and `TestWebhookIgnoreJID_ResolverErrorStillForwardsNonGroupEvent`
- **atomic preserve**: added `WebhookIgnoreGroupsSet` to `DeviceWebhookConfig`; `SetDeviceWebhookConfig` does `webhook_ignore_groups = CASE WHEN ? THEN ? ELSE webhook_ignore_groups END` in the single UPDATE, so the handler no longer pre-reads the row. it only reads *after* the write, and only to report the stored value when the field was omitted. test: `TestSQLiteRepositoryDeviceWebhookConfig_PreserveIsAtomicUnderConcurrency` (25 concurrent omitted-field writes vs 1 explicit write, explicit always wins — i checked it fails against the old unconditional overwrite). sqlite is the only backend in the repo so nothing else to mirror

src/infrastructure/whatsapp/webhook_forward.go
```diff
@@ -99,16 +99,23 @@ func getContactMutex(phone string) *sync.Mutex {
 // successful targets still receive the event.
 func forwardPayloadToConfiguredWebhooks(ctx context.Context, payload map[string]any, eventName string) error {
 	deviceJID, _ := payload["device_id"].(string)
-	webhookConfig, err := getWebhookConfigForDevice(deviceJID)
-	if err != nil {
+	record, err := resolveWebhookDeviceRecord(ctx, payload)
+	recordResolutionFailed := err != nil
```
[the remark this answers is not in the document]
Kizuno18: not applying this one. `(nil, nil)` here is not an unknown policy, it's "this device has no record", and that's the normal state for every payload in the existing test suite and for any device whose record hasn't been written yet — failing closed on it would drop every group webhook for those devices. the ambiguity case from the original review (two sibling slots on one number) is handled by the AD-JID-first branch in `resolveWebhookDeviceRecord`: the exact slot resolves before the bare-number fallback ever runs. the fail-closed path stays scoped to a real lookup error, where the record may exist but can't be read

src/ui/rest/device.go
Kizuno18: Fixed through 21bfc63. WebhookIgnoreGroupsSet carries field presence, and the repository preserves omitted values in a single UPDATE with CASE. The regression races omitted-field updates against an explicit false write.

src/infrastructure/whatsapp/webhook_forward.go
Kizuno18: Fixed the remaining (nil, nil) case in 118cd11. Group forwarding now fails closed for lookup errors and missing device records. Added a forwarding regression that failed on the previous head; the affected package and full Go suite pass.

PR #671, opened 2026-05-08
feat: add per-device webhook support

mohamedhabibwork:
## Summary

Add per-device webhook support where each device can have its own webhook URL. If a device has no custom webhook URL, it falls back to the global webhook configuration.

## Flow Diagram

```mermaid
sequenceDiagram
    participant WA as WhatsApp Event
    participant EH as Event Handler
    participant WF as Webhook Forwarder
    participant DB as Device DB
    participant WH as Webhook URL

    WA->>EH: Event received
    EH->>WF: forwardPayloadToConfiguredWebhooks(payload)

    rect rgba(150, 200, 255, 0.3)
    Note over WF: STEP 1: Extract device_id from payload
    WF->>DB: GetDeviceRecordByJID(deviceJID)
    DB-->>WF: DeviceRecord {webhook_url}
    end

    alt device has custom webhook_url
        WF->>WF: Use device webhook (override global)
    else device webhook is null/empty
        WF->>WH: Use global config.WhatsappWebhook
    end

    WF->>WF: forwardToWebhooks() with HMAC + retries
```

## API Endpoints

- `PATCH /devices/{device_id}/webhook` - Set device-specific webhook URL
  - Body: { "webhook_url": "https://example.com/webhook" }
  - Set to empty string "" to clear and use global webhook

- `GET /devices/{device_id}/webhook` - Get device-specific webhook URL
  - Returns empty string if not set (use global)

## Changes

- Add `webhook_url` column to `devices` table (migration #17)
- Add device-specific webhook override logic in webhook forwarding
- Add REST API endpoints for getting/setting device webhook URL
- Add unit tests for webhook forwarding and device service
- Update OpenAPI documentation

## Behavior

1. If device has a custom webhook URL => use it (override global)
2. If device webhook is null/empty => use global `config.WhatsappWebhook`

## Test Results

- `cd src && go test ./...` - All tests pass
- `cd src && go vet ./...` - No issues
- `cd src && go build` - Builds successfully
- Manual testing via PATCH/GET endpoints confirmed webhook URL is stored/retrieved correctly
- Webhook forwarding correctly uses device webhook when set, falls back to global when cleared

PR #833, opened 2026-09-05
feat(device): add per-device chat storage and auto-download-media overrides

Kizuno18:
## Context

- Add per-device `chat_storage` and `auto_download_media` overrides, following the per-device webhook mechanism from #671, exposed as `GET`/`PATCH /devices/{device_id}/settings`.
- Motivation: when one instance hosts many devices, chat storage and `--auto-download-media` are all-or-nothing. Some devices (notification-only accounts, high-volume media chats) should be lighter without changing behavior for every other device on the instance.
- Defaults are unchanged. Both columns are nullable and `NULL` means "no override": chat storage stays always-on (there is no global chat-storage flag since the SQLite migration), and `auto_download_media` keeps following `--auto-download-media` / `WHATSAPP_AUTO_DOWNLOAD_MEDIA`. A device-level `true`/`false` wins over the global flag only when explicitly set.
- `PATCH` is a partial update: omitted fields are left untouched, a field sent as `null` clears the override, and at least one field is required. The body is parsed as raw JSON so "absent" and "null" can be told apart (a `*bool` struct field decodes to `nil` in both cases).
- `chat_storage=false` gates the inbound path only: `CreateMessage`/`CreateReaction` in `event_message_handler.go`, `CreateIncomingCallRecord` in `event_call.go`, the auto-reply's own sent-message record in `auto_reply.go`, and history sync backfill. Outbound writes from the send/edit/delete use cases are intentionally still recorded, since those are actions the API caller asked for.
- `auto_download_media` replaces the direct `config.WhatsappAutoDownloadMedia` checks in `event_message.go` / `event_message_handler.go` with a per-client resolver that falls back to the global flag.
- Lookup is uncached, same as `getWebhookConfigForDevice`; `GetDeviceRecordByJID` now returns the two new columns alongside the webhook ones it already selected.
- Migrations 45 and 46 add `devices.chat_storage` and `devices.auto_download_media` (`BOOLEAN DEFAULT NULL`).
- Documented in `readme.md` and `docs/openapi.yaml`. No UI change: the embedded device manager was removed in #766, so this only adds the API for `gowa-ui` to wire up later.

### API

```
GET   /devices/{device_id}/settings
PATCH /devices/{device_id}/settings   {"chat_storage": false, "auto_download_media": true}
PATCH /devices/{device_id}/settings   {"chat_storage": null}   # clear one override, leave the other alone
```

Response `results`: `{"device_id": "...", "chat_storage": bool|null, "auto_download_media": bool|null}`.

## Test Results

- `go build ./...`, `go vet ./...` clean.
- `go test ./... -count=1`: all 13 packages `ok`.
- 24 new tests:
  - `infrastructure/chatstorage` (7): schema round-trip, partial update leaves the other field untouched, explicit `nil` clears vs empty patch is a no-op, unknown device, `GetDeviceRecordByJID` surfaces the new columns.
  - `infrastructure/whatsapp` (11): resolver rules for both flags — no device id, device not found, no override, explicit true/false, override wins over the global flag, lookup error falls back to the default.
  - `ui/rest` (6): PATCH with both fields, explicit `null` vs absent field, empty body rejected, non-bool value rejected, GET with and without an override.

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -978,6 +978,7 @@ paths:
           application/json:
             schema:
               type: object
+              minProperties: 1
               properties:
                 chat_storage:
                   type: boolean
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -979,6 +979,7 @@ paths:
             schema:
               type: object
               minProperties: 1
+              additionalProperties: false
               properties:
                 chat_storage:
                   type: boolean
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -887,6 +887,160 @@ paths:
               schema:
                 $ref: '#/components/schemas/ErrorInternalServer'
 
+  /devices/{device_id}/settings:
+    get:
+      operationId: getDeviceStorageSettings
+      tags:
+        - device
+      summary: Get device storage settings
+      description: |
+        Get the per-device chat_storage / auto_download_media overrides for a specific device.
+        A null value means the device has no override and follows the instance-wide default
+        (chat storage is always on; auto_download_media follows --auto-download-media /
+        WHATSAPP_AUTO_DOWNLOAD_MEDIA).
+      parameters:
+        - name: device_id
+          in: path
+          required: true
+          schema:
+            type: string
+          description: Device ID
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  status:
+                    type: integer
+                    example: 200
+                  code:
+                    type: string
+                    example: SUCCESS
+                  message:
+                    type: string
+                    example: Device storage settings retrieved
+                  results:
+                    type: object
+                    properties:
+                      device_id:
+                        type: string
+                        description: The device ID
+                        example: 'my-device'
+                      chat_storage:
+                        type: boolean
+                        nullable: true
+                        description: Per-device chat storage override. Null follows the instance-wide default (always on).
+                        example: false
+                      auto_download_media:
+                        type: boolean
+                        nullable: true
+                        description: Per-device auto-download-media override. Null follows --auto-download-media / WHATSAPP_AUTO_DOWNLOAD_MEDIA.
+                        example: true
+        '404':
+          description: Device not found
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorNotFound'
+        '500':
+          description: Internal Server Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorInternalServer'
+    patch:
+      operationId: updateDeviceStorageSettings
+      tags:
+        - device
+      summary: Set device storage settings
+      description: |
+        Apply a partial update to a device's chat_storage / auto_download_media overrides.
+        Only the fields present in the request body are changed — an omitted field keeps its
+        current value. A field sent as `null` clears the override, falling back to the
+        instance-wide default. At least one of chat_storage or auto_download_media is required.
+
+        Multi-tenant deployments running one instance for many customer devices use this to
+        keep specific devices (e.g. alarm/service accounts) out of chat storage or off
+        automatic media downloads without changing the instance-wide flags for every other
+        device.
+      parameters:
+        - name: device_id
+          in: path
+          required: true
+          schema:
+            type: string
+          description: Device ID
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                chat_storage:
+                  type: boolean
+                  nullable: true
+                  description: Set to enable/disable chat storage for this device, or null to clear the override.
+                  example: false
+                auto_download_media:
+                  type: boolean
+                  nullable: true
+                  description: Set to enable/disable auto media download for this device, or null to clear the override.
+                  example: true
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  status:
+                    type: integer
+                    example: 200
+                  code:
+                    type: string
+                    example: SUCCESS
+                  message:
+                    type: string
+                    example: Device storage settings updated
+                  results:
+                    type: object
+                    properties:
+                      device_id:
+                        type: string
+                        description: The device ID
+                        example: 'my-device'
+                      chat_storage:
+                        type: boolean
+                        nullable: true
+                        example: false
+                      auto_download_media:
+                        type: boolean
+                        nullable: true
+                        example: true
+        '400':
+          description: Bad Request
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorBadRequest'
+        '404':
+          description: Device not found
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorNotFound'
+        '500':
+          description: Internal Server Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorInternalServer'
+
   /user/info:
     get:
       operationId: userInfo
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -962,10 +962,9 @@ paths:
         current value. A field sent as `null` clears the override, falling back to the
         instance-wide default. At least one of chat_storage or auto_download_media is required.
 
-        Multi-tenant deployments running one instance for many customer devices use this to
-        keep specific devices (e.g. alarm/service accounts) out of chat storage or off
-        automatic media downloads without changing the instance-wide flags for every other
-        device.
+        Useful when one instance hosts many devices: specific devices (e.g. notification-only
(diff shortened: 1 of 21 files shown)


src/infrastructure/whatsapp/device_storage_settings.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	return nil, nil
+}
+
+// clientDeviceJID returns the own-account JID (bare number form) for a connected
+// client, or "" when the client or its store is not yet available.
+func clientDeviceJID(client *whatsmeow.Client) string {
+	if client == nil || client.Store == nil || client.Store.ID == nil {
+		return ""
+	}
```
aldinokemal: **P1 — preserve the exact device/slot identity when resolving these overrides.** `client.Store.ID` is the companion AD JID, but `ToNonAD()` removes the device component before `GetDeviceRecordByJID`. Current main deliberately treats a bare-number lookup as ambiguous when two slots are sibling companions for the same account and returns `nil`; only the full AD JID resolves the exact slot. That means `slot-a: chat_storage=false` and `slot-b: chat_storage=true` can both fall through to the default instead of honoring their own settings. The exact event `DeviceInstance` is already present in context, so prefer its device ID / exact record (or at minimum the full AD JID) rather than discarding the differentiating identity. Please add a regression test with two slots sharing the same non-AD JID and opposite overrides.

src/infrastructure/whatsapp/device_storage_settings.go
```diff
@@ -0,0 +1,94 @@
...
+// client/store.
+func isChatStorageEnabledForDeviceJID(deviceJID string) bool {
+	if deviceJID == "" {
+		return true
+	}
+
+	record, err := getDeviceRecordForStorageSettings(deviceJID)
+	if err != nil {
+		logrus.Warnf("Failed to resolve chat_storage override for device %s, defaulting to enabled: %v", deviceJID, err)
+		return true
```
aldinokemal: **P1 — a policy lookup error must not silently enable an explicit opt-out.** On a transient repository error we cannot distinguish “no override” from a persisted `chat_storage=false`, yet this path returns `true`; the analogous auto-download path falls back to the global flag, which may be `true` while the device override is `false`. A temporary storage error can therefore persist incoming plaintext or download media contrary to the device policy. Fail closed for an indeterminate lookup (or use a safely maintained last-known policy), while keeping the normal default only for a successful lookup with no override. Please update the lookup-error tests and cover the handler/media side effects. The same issue applies to the auto-download error branch below.

src/infrastructure/whatsapp/event_message_handler.go
```diff
@@ -34,18 +34,24 @@ func handleMessage(ctx context.Context, evt *events.Message, chatStorageRepo dom
 	evt = materializeSecretEditMessage(ctx, evt, client)
 	pollPayload := preparePollWebhookPayload(ctx, client, chatStorageRepo, evt)
 
+	chatStorageEnabled := isChatStorageEnabledForClient(client)
```
aldinokemal: **P1 — apply the chat-storage policy before poll persistence.** `preparePollWebhookPayload` runs before `chatStorageEnabled` is checked. It can persist poll definitions for creation/edit events and append options for add-option events, so `chat_storage=false` still writes poll content even though message/reaction storage is skipped. Compute the policy first and ensure poll preparation cannot mutate chat storage when disabled (while preserving any read/decrypt work needed purely for webhook payloads). Please add disabled-storage regression coverage for poll creation, edit, and add-option.

aldinokemal (changes requested):
Blocking per-device policy issues remain on this head. The repository partial-update implementation, append-only migrations, interface/wrapper wiring, and additive REST API are generally sound, but the runtime policy path loses exact companion identity, fails open when the policy lookup errors, and allows poll persistence before the chat-storage gate. Those paths can persist or download data despite an explicit per-device false override. Please address the inline findings and add regression coverage. There is no PR-triggered GitHub Actions run for this head; the published commit status is CodeRabbit only.

Kizuno18:
pushed 3 commits on top

- **exact slot identity**: dropped the `client.Store.ID.ToNonAD()` + `GetDeviceRecordByJID` lookup. the policy now comes from the `DeviceInstance` already in ctx (`instance.ID()`) via `GetDeviceStorageSettings(deviceID)`, the same device-ID-keyed lookup the REST GET/PATCH use, so sibling slots on one number never collapse into each other. test: `TestIsChatStorageEnabledForClient_SiblingSlotsShareNumberButHonorOwnOverride`
- **fail closed**: both `isChatStorageEnabledForDeviceID` and `isAutoDownloadMediaEnabledForDeviceID` return false on a lookup error. default-enabled only for a successful lookup with no override. tests: `TestIsChatStorageEnabledForDeviceID_LookupErrorFailsClosed`, `TestIsAutoDownloadMediaEnabledForDeviceID_LookupErrorFailsClosed`
- **polls**: `chatStorageEnabled` is computed before `preparePollWebhookPayload` and threaded into it, gating `UpsertPollDefinition` (create/edit) and `AppendPollOption`; decrypt/read work for the webhook payload is unchanged. tests: `TestPreparePollWebhookPayload{Creation,Edit,AddOption}SkipsPersistenceWhenChatStorageDisabled`
- also: `minProperties: 1` on the patch schema in openapi (coderabbit), skip the `DEVICE_STORAGE_SETTINGS_UPDATED` broadcast on an empty patch and fixed the `(nil, nil)` doc on the interface (kilo)

docs/openapi.yaml
```diff
@@ -887,6 +887,160 @@ paths:
...
+          schema:
+            type: string
+          description: Device ID
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              minProperties: 1
```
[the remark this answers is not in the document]
Kizuno18: added `additionalProperties: false` in 8395f63

Kizuno18:
on the out-of-diff `pollDeviceID` note from coderabbit: that function and the JID-keyed poll definitions are current main behavior, not something this PR introduces or touches (this PR only threads the storage policy into `preparePollWebhookPayload`). re-keying poll definitions to `instance.ID()` needs a migration/lookup fallback for existing rows, so it belongs in its own PR. happy to open one after this lands if you want it

src/usecase/device.go
```diff
@@ -315,6 +315,65 @@ func (s *serviceDevice) GetDeviceWebhookConfig(ctx context.Context, deviceID str
...
+	storage := s.manager.GetStorage()
+	if storage == nil {
+		return fmt.Errorf("storage not available")
+	}
+
+	if err := storage.SetDeviceStorageSettings(deviceID, patch); err != nil {
+		return fmt.Errorf("failed to set device storage settings: %w", err)
+	}
+
+	websocket.Broadcast <- websocket.BroadcastMessage{
```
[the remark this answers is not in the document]
Kizuno18: done in 4ebed24: the usecase returns before the broadcast when neither field is flagged present

src/domains/chatstorage/interfaces.go
```diff
@@ -119,6 +119,13 @@ type IChatStorageRepository interface {
 	SetDeviceWebhookConfig(deviceID string, config *DeviceWebhookConfig) error
 	// GetDeviceWebhookConfig retrieves the full webhook configuration for a device.
 	GetDeviceWebhookConfig(deviceID string) (*DeviceWebhookConfig, error)
+	// SetDeviceStorageSettings applies a partial update to the per-device
+	// chat_storage / auto_download_media overrides. Only the fields flagged as
+	// present in patch are touched.
+	SetDeviceStorageSettings(deviceID string, patch DeviceStoragePatch) error
+	// GetDeviceStorageSettings retrieves the per-device chat_storage /
+	// auto_download_media overrides. Returns nil fields when no override is set.
```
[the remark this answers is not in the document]
Kizuno18: reworded in 4ebed24: `(nil, nil)` only for a missing device, existing devices always get a non-nil struct with possibly-nil fields

src/infrastructure/whatsapp/device_storage_settings.go
Kizuno18: Fixed in 0bd5351 and 4ebed24. Policy lookup now uses the exact DeviceInstance.ID(). The sibling-slot regression checks opposite overrides for two slots sharing the same account number.

src/infrastructure/whatsapp/event_message_handler.go
Kizuno18: Fixed in 0bd5351 and 4ebed24. The storage policy is resolved before poll preparation and gates creation, edit, and add-option writes. Disabled-storage regressions cover all three paths.

src/infrastructure/whatsapp/device_storage_settings.go
Kizuno18: Added the missing handler coverage in af86b67. The regression sends text, reaction, and image events through handleMessage with a real device context and a failing settings lookup, then checks that CreateMessage, CreateReaction, and media extraction are never called, even with global auto-download enabled. The WhatsApp package and the race-enabled regression pass.

PR #834, opened 2026-09-06
feat(chat): add on-demand history sync endpoint

Kizuno18:
## Context

- Add `POST /chat/{chat_jid}/history` so a client can ask the phone for older messages of a chat on demand — the same thing WhatsApp Web does when you scroll up ("load older messages"). whatsmeow already exposes this via `Client.BuildHistorySyncRequest(lastKnownMessageInfo, count)` + `Client.SendPeerMessage`, this just wires it into the REST API on top of the existing chat storage.
- The request is anchored at the oldest message currently stored locally for that chat (`GetOldestMessageByDevice`, new on `IChatStorageRepository`), since the phone needs an oldest known message ID/timestamp to know where to start looking for older history. A chat with nothing stored yet returns `400` instead of sending an unanchored request.
- The endpoint only confirms the request was sent (`status: "requested"` + the anchor used); the phone answers asynchronously with a `HISTORY_SYNC_ON_DEMAND` history sync event, which gets persisted the same way the initial bootstrap does.
- Along the way, fixed a bug in `processHistorySync`: the sync-type switch only handled `INITIAL_BOOTSTRAP`/`RECENT`/`PUSH_NAME`, so an on-demand reply from the phone would hit the `default` branch and be silently dropped instead of stored. `HISTORY_SYNC_ON_DEMAND` carries conversations in the same shape as `RECENT`, so it now routes through the same `processConversationMessages`.
- Documented in `readme.md` and `docs/openapi.yaml`.

### API

```
POST /chat/{chat_jid}/history   {"count": 50}
```

- `count` is optional, defaults to 50, capped at 500.
- `results`: `{"status": "requested", "chat_jid": "...", "requested_count": 50, "anchor_message_id": "...", "anchor_timestamp": "..."}`.
- `400` when `chat_jid` is missing, `count` is out of range, or the chat has no stored messages to anchor on.

## Test Results

- `go build ./...`, `go vet ./...` clean.
- `go test ./... -count=1`: all packages `ok` (1554 tests, up from 1542 on `main`).
- 12 new tests:
  - `validations` (6): default count, cap, min/max bounds, blank `chat_jid`.
  - `usecase` (3): validation short-circuits before touching storage, missing client returns `ErrWaCLI`, chat with no stored messages returns a `ValidationError` without requiring a live connection.
  - `infrastructure/chatstorage` (1): `GetOldestMessageByDevice` returns the earliest message by timestamp, scoped by device, and `nil` for an empty chat.
  - `infrastructure/whatsapp` (1): `HISTORY_SYNC_ON_DEMAND` is routed to `processConversationMessages` and actually persisted (this test fails against `main`, confirming the routing bug it fixes) — plus the pre-existing 1542 tests all still pass unmodified.

### DIFF
--- a/src/infrastructure/chatstorage/sqlite_repository_message_lookup.go
+++ b/src/infrastructure/chatstorage/sqlite_repository_message_lookup.go
@@ -26,3 +26,26 @@ func (r *SQLiteRepository) GetMessageByIDChatAndDevice(deviceID, chatJID, id str
 	}
 	return message, err
 }
+
+// GetOldestMessageByDevice returns the earliest stored message for a chat,
+// scoped to the device for data isolation. Used to anchor on-demand history
+// sync requests, which need the oldest known message ID/timestamp to ask the
+// phone for anything older. Returns (nil, nil) when the chat has no stored
+// messages yet.
+func (r *SQLiteRepository) GetOldestMessageByDevice(deviceID, chatJID string) (*domainChatStorage.Message, error) {
+	query := `
+		SELECT id, chat_jid, device_id, sender, content, timestamp, is_from_me,
+			media_type, call_metadata, filename, url, direct_path, media_key, file_sha256,
+			file_enc_sha256, file_length, referral_metadata, created_at, updated_at
+		FROM messages
+		WHERE chat_jid = ? AND device_id = ?
+		ORDER BY timestamp ASC
+		LIMIT 1
+	`
+
+	message, err := r.scanMessage(r.db.QueryRow(query, chatJID, deviceID))
+	if err == sql.ErrNoRows {
+		return nil, nil
+	}
+	return message, err
+}
--- a/src/infrastructure/chatstorage/sqlite_repository_message_lookup.go
+++ b/src/infrastructure/chatstorage/sqlite_repository_message_lookup.go
@@ -32,13 +32,18 @@ func (r *SQLiteRepository) GetMessageByIDChatAndDevice(deviceID, chatJID, id str
 // sync requests, which need the oldest known message ID/timestamp to ask the
 // phone for anything older. Returns (nil, nil) when the chat has no stored
 // messages yet.
+//
+// Locally synthesized rows (e.g. incoming calls, stored as media_type "call"
+// with id "call:<callID>") are excluded: the phone only recognizes real
+// WebMessageInfo IDs as history-sync anchors, so a call row would repeatedly
+// produce an invalid oldest_msg_id.
 func (r *SQLiteRepository) GetOldestMessageByDevice(deviceID, chatJID string) (*domainChatStorage.Message, error) {
 	query := `
 		SELECT id, chat_jid, device_id, sender, content, timestamp, is_from_me,
 			media_type, call_metadata, filename, url, direct_path, media_key, file_sha256,
 			file_enc_sha256, file_length, referral_metadata, created_at, updated_at
 		FROM messages
-		WHERE chat_jid = ? AND device_id = ?
+		WHERE chat_jid = ? AND device_id = ? AND (media_type IS NULL OR media_type != 'call')
 		ORDER BY timestamp ASC
 		LIMIT 1
 	`
--- a/src/infrastructure/whatsapp/history_sync.go
+++ b/src/infrastructure/whatsapp/history_sync.go
@@ -67,8 +67,11 @@ func processHistorySync(ctx context.Context, data *waHistorySync.HistorySync, ch
 	log.Infof("Processing history sync type: %s", syncType.String())
 
 	switch syncType {
-	case waHistorySync.HistorySync_INITIAL_BOOTSTRAP, waHistorySync.HistorySync_RECENT:
-		// Process conversation messages
+	case waHistorySync.HistorySync_INITIAL_BOOTSTRAP, waHistorySync.HistorySync_RECENT, waHistorySync.HistorySync_ON_DEMAND:
+		// Process conversation messages. ON_DEMAND is the phone's reply to a
+		// history sync request built with Client.BuildHistorySyncRequest (older
+		// messages fetched on demand, e.g. "load older messages"); it carries
+		// conversations in the same shape as INITIAL_BOOTSTRAP/RECENT.
 		return processConversationMessages(ctx, data, chatStorageRepo, client)
 	case waHistorySync.HistorySync_PUSH_NAME:
 		// Process push names to update chat names
--- a/src/infrastructure/whatsapp/history_sync.go
+++ b/src/infrastructure/whatsapp/history_sync.go
@@ -294,6 +294,20 @@ func processConversationMessages(ctx context.Context, data *waHistorySync.Histor
 				EphemeralExpiration: ephemeralExpiration,
 			}
 
+			if data.GetSyncType() == waHistorySync.HistorySync_ON_DEMAND {
+				// ON_DEMAND delivers messages older than the local anchor, so its
+				// batch timestamp must never move the chat's activity time backward
+				// or clobber flags (e.g. archived) that live history didn't touch.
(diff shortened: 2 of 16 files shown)


src/infrastructure/whatsapp/history_sync.go
```diff
@@ -67,8 +67,11 @@ func processHistorySync(ctx context.Context, data *waHistorySync.HistorySync, ch
 	log.Infof("Processing history sync type: %s", syncType.String())
 
 	switch syncType {
-	case waHistorySync.HistorySync_INITIAL_BOOTSTRAP, waHistorySync.HistorySync_RECENT:
-		// Process conversation messages
+	case waHistorySync.HistorySync_INITIAL_BOOTSTRAP, waHistorySync.HistorySync_RECENT, waHistorySync.HistorySync_ON_DEMAND:
```
aldinokemal, line 70: **P1 — ON_DEMAND must not rewrite current chat state from an older batch.** At the pinned whatsmeow revision, this sync type contains messages immediately *before* the anchor. Routing it unchanged through `processConversationMessages` later builds a new `Chat` whose `LastMessageTime` is the newest timestamp in that older batch and whose `Archived` field is the zero value. Current `StoreChat` unconditionally updates both `last_message_time` and `archived`, so loading older history can move an active chat backwards in the chat list **and silently unarchive an archived chat**. Please preserve the existing device-scoped chat metadata for ON_DEMAND (at minimum max/current `LastMessageTime` and `Archived`; ideally separate older-message ingestion from live chat-metadata updates). Add a regression test with an existing newer `LastMessageTime` and `Archived=true` and verify both survive an ON_DEMAND response.

src/ui/rest/chat.go
```diff
@@ -205,6 +206,42 @@ func (controller *Chat) ArchiveChat(c fiber.Ctx) error {
...
+			Status:  400,
+			Code:    "BAD_REQUEST",
+			Message: "invalid chat_jid path parameter: " + err.Error(),
+			Results: nil,
+		})
+	}
+	request.ChatJID = chatJID
+
+	// Parse JSON body (optional — count defaults when the body is empty)
+	if err := c.Bind().Body(&request); err != nil {
```
aldinokemal: **P1 — the optional body is not actually optional, and it can override the path JID.** With the pinned Fiber v3.4.0, `Bind().Body()` selects a binder from `Content-Type` and returns `ErrUnprocessableEntity` when no supported type is present, so a bodyless POST with no content type returns 400 instead of using the default count. Also, `request.ChatJID` is assigned from the route *before* binding, while the struct exposes `json:"chat_jid"`; a body such as `{"chat_jid":"other@s.whatsapp.net","count":50}` therefore replaces the authoritative route value before storage lookup / `BuildHistorySyncRequest`. Bind only when a body exists (prefer a body DTO containing only `count`), then assign the decoded route JID after binding. Please add REST tests for an empty/no-Content-Type body and a conflicting body `chat_jid`.

src/infrastructure/chatstorage/sqlite_repository_message_lookup.go
```diff
@@ -26,3 +26,26 @@ func (r *SQLiteRepository) GetMessageByIDChatAndDevice(deviceID, chatJID, id str
...
+// sync requests, which need the oldest known message ID/timestamp to ask the
+// phone for anything older. Returns (nil, nil) when the chat has no stored
+// messages yet.
+func (r *SQLiteRepository) GetOldestMessageByDevice(deviceID, chatJID string) (*domainChatStorage.Message, error) {
+	query := `
+		SELECT id, chat_jid, device_id, sender, content, timestamp, is_from_me,
+			media_type, call_metadata, filename, url, direct_path, media_key, file_sha256,
+			file_enc_sha256, file_length, referral_metadata, created_at, updated_at
+		FROM messages
+		WHERE chat_jid = ? AND device_id = ?
```
aldinokemal: **P1 — only use real WhatsApp message rows as history anchors.** Current main stores incoming calls as synthetic rows in this same table with `ID = "call:" + evt.CallID` and `MediaType = "call"`. This query can therefore select a call row as the oldest item. The exact pinned `BuildHistorySyncRequest` forwards the selected `ID`, timestamp and from-me flag as the primary-device `oldest_msg_*` anchor; the phone has no chat message with our locally invented `call:<id>` ID, so a chat whose oldest row is a call can repeatedly send an invalid/non-progressing anchor. Exclude synthetic call rows (and any other locally synthesized IDs that cannot exist as `WebMessageInfo` anchors) and add a test where an older call row plus a newer real message returns the real message.

aldinokemal (changes requested):
The protocol direction is correct against the pinned whatsmeow revision (`28bfe537ea6a`): `BuildHistorySyncRequest` is intended for `SendPeerMessage`, returns an `ON_DEMAND` history-sync response, and uses the anchor chat/ID/from-me/timestamp. However, I found three blocking correctness/data-integrity issues in the current integration: ON_DEMAND currently mutates live chat metadata using older history, the REST handler does not actually support an empty optional body and allows JSON to override the route JID, and the anchor query can select locally-synthetic call rows that the phone cannot recognize as WhatsApp message IDs. Please address the inline findings and add the requested regression coverage before this is merged.

Kizuno18:
pushed 302e4c0, 3798841, 806ee37, c9d246b

- **on-demand chat metadata**: `processConversationMessages` now loads the existing device-scoped chat when `sync_type == ON_DEMAND` and keeps `max(existing, batch)` for `LastMessageTime` plus the stored `Archived`. if that lookup errors it skips the chat upsert entirely (messages still get stored) instead of guessing. test: `TestProcessConversationMessagesOnDemandPreservesNewerChatMetadata`
- **optional body / route jid**: body is bound only when `len(c.Body()) > 0`, and `request.ChatJID` is assigned from the route *after* binding so a body `chat_jid` can't override it. tests: `TestRequestChatHistoryAllowsEmptyBody`, `TestRequestChatHistoryIgnoresConflictingBodyChatJID`
- **anchor**: `GetOldestMessageByDevice` excludes `media_type = 'call'` rows (NULL-safe). no postgres variant of this query in the repo. test: `TestSQLiteRepositoryGetOldestMessageByDeviceSkipsSyntheticCallRows`

src/ui/rest/chat_test.go
```diff
@@ -76,3 +79,75 @@ func TestChatJIDParamRejectsMalformedEscape(t *testing.T) {
...
+// optional body must not require a Content-Type to fall back to defaults.
+func TestRequestChatHistoryAllowsEmptyBody(t *testing.T) {
+	app := fiber.New()
+	spy := &requestChatHistorySpy{}
+	controller := &Chat{Service: spy}
+	app.Post("/chat/:chat_jid/history", controller.RequestChatHistory)
+
+	req := httptest.NewRequest("POST", "/chat/6289685028129%40s.whatsapp.net/history", nil)
+
+	resp, err := app.Test(req)
```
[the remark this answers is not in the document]
Kizuno18: done in d27fae0, all three `app.Test` responses in the file close their body now

aldinokemal (approved):
Re-review on head d27fae0a09b8ac834c578b53a5504c58148fc2a3. All three prior owner blockers are fixed: ON_DEMAND preserves newer device-scoped chat activity/archive state, the optional REST body no longer fails bodyless requests or overrides the route JID, and synthetic call rows are excluded from history anchors. The protocol flow remains consistent with the pinned whatsmeow 28bfe537ea6a BuildHistorySyncRequest/SendPeerMessage contract. No new blocking correctness, protocol, device-isolation, or API-contract issue found. Safe to merge from review perspective.

Kizuno18:
@coderabbitai @aldinokemal safe to merge? could you merge/squash?

penard-monkey:
Hi folks. Any idea when merge will happen? Really eager to get this feature. :)

Kizuno18:
@penard-monkey No ETA from my side yet. The requested fixes are in and the maintainer has approved the latest revision. It's waiting for merge.

PR #836, opened 2026-09-07
feat: add flag to ignore downloading status media

danihidayatx:
## Description
This PR introduces the `WHATSAPP_IGNORE_STATUS_MEDIA` (or `--ignore-status-media`) configuration flag to solve the issue of WhatsApp status media (`status@broadcast`) being automatically downloaded when the `WHATSAPP_AUTO_DOWNLOAD_MEDIA` flag is active.

## Changes Made
- Added `WhatsappIgnoreStatusMedia` boolean config to `settings.go`.
- Plumbed the configuration through Viper and Cobra in `root.go` as `--ignore-status-media`.
- Added an explicit guard clause in `event_message_handler.go`'s `handleImageMessage` to skip media download for status updates if the flag is enabled.
- Updated documentation in `readme.md` and `src/.env.example`.

## Impact
- **Default behavior:** Unchanged (`false`). By default, status media will still be downloaded if auto-download is active.
- **Opt-in behavior:** When set to `true`, images originating from `status@broadcast` are skipped during the download flow.


src/infrastructure/whatsapp/event_message_handler_test.go
```diff
@@ -289,3 +289,55 @@ func reactionEventForTest(eventID, targetID, emoji string) *events.Message {
...
+			Info: types.MessageInfo{
+				Chat: types.NewJID("123456789", types.BroadcastServer),
+			},
+			Message: &waE2E.Message{
+				ImageMessage: &waE2E.ImageMessage{},
+			},
+		}
+		
+		defer func() {
+			if r := recover(); r == nil {
```
aldinokemal: **Blocker — this test's panic spy is not valid with the pinned whatsmeow.** The test message has an empty `ImageMessage`, so `Client.Download` sees no `DirectPath` and returns `ErrNoURLPresent` before accessing any uninitialized client internals. `utils.ExtractMedia` returns that error and `handleImageMessage` just logs it, so no panic occurs; this deferred check then fails the `does not skip normal broadcast list` subtest even when the production guard is correct. Please assert the branch through an injectable downloader/pure predicate (or another deterministic seam) rather than expecting a panic.

aldinokemal (changes requested):
The production guard itself is pointed at the right identity: current `src/go.mod` pins whatsmeow `28bfe537ea6a`, where `types.StatusBroadcastJID` is exactly `status@broadcast`, distinct from ordinary broadcast-list JIDs. The config/CLI/env wiring is additive and follows the existing auto-download pattern.

One blocker remains in the added regression test: the "normal broadcast list" branch cannot use a panic as a spy. With the empty `ImageMessage{}` in this test, pinned whatsmeow's `Client.Download` returns `ErrNoURLPresent` before touching client internals, `utils.ExtractMedia` propagates that error, and `handleImageMessage` only logs it. Therefore this subtest reaches its deferred recover with `r == nil` and fails every time even though the guard correctly allowed the message through.

Please make the behavior observable without relying on an implementation panic (for example inject/wrap the media downloader, or test a small pure `shouldSkipStatusMedia` predicate/table) and run the exact package test plus `go test ./...`. I found no separate runtime blocker in the status-JID comparison itself.

src/infrastructure/whatsapp/event_message_handler_test.go
danihidayatx: Thanks for the catch! You're completely right about Client.Download returning ErrNoURLPresent early preventing the panic.

I have refactored the logic into a small pure predicate function (shouldIgnoreImageDownload) just as you suggested. The panic-based spy test has been completely removed and replaced with a clean table-driven test (TestShouldIgnoreImageDownload) that deterministically verifies all the JID combinations without touching the client internals.

Let me know if this looks good to you!

aldinokemal (approved):
Re-review on the current head confirms the blocking test issue is fixed. The implementation now uses a deterministic pure predicate instead of relying on a panic from an empty media message, and the table-driven regression covers status@broadcast, ordinary broadcast-list JIDs, normal user JIDs, the global auto-download disable, and the opt-out-disabled case. Against the exact pinned whatsmeow 28bfe537ea6a, `types.StatusBroadcastJID` is exactly `status@broadcast` and ordinary broadcast lists are distinct, so the JID comparison is protocol-correct. The env/Viper/Cobra plumbing follows the existing auto-download configuration pattern and defaults preserve current behavior. I found no new correctness, protocol, API, security, or compatibility blocker. No PR-triggered GitHub Actions run is present for this exact head.

src/infrastructure/whatsapp/event_message_handler.go
```diff
@@ -82,6 +82,9 @@ func handleImageMessage(ctx context.Context, evt *events.Message, client *whatsm
 	if !config.WhatsappAutoDownloadMedia {
 		return
 	}
+	if config.WhatsappIgnoreStatusMedia && strings.Contains(evt.Info.SourceString(), "broadcast") {
```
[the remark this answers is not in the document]
aldinokemal: Resolving as obsolete: this describes a revision that no longer exists on the branch.

The guard was rewritten in `a44ad51` and now uses an exact JID comparison, not a substring match:

```go
if ignoreStatusMedia && chatJID == types.StatusBroadcastJID {
```

whatsmeow defines broadcast lists as disjoint from status (`types/jid.go`):

```go
func (jid JID) IsBroadcastList() bool {
	return jid.Server == BroadcastServer && jid.User != StatusBroadcastJID.User
}
```

So `1234567890@broadcast` is `JID{User:"1234567890", Server:"broadcast"}`, which is not `==` to `JID{User:"status", Server:"broadcast"}`. The false positive is structurally impossible, and `TestShouldIgnoreImageDownload/normal_broadcast_list_not_ignored_when_status_ignored` covers exactly that case.

PR #840, opened 2026-09-10
fix(mcp): mount GET /mcp for streamable-HTTP client handshake

### LINKED ISSUE #839: MCP transport: GET /mcp returns 405 → unusable with Claude Code, Cursor, and other streamable-HTTP clients
## Summary

The streamable-HTTP MCP endpoint (`/mcp`) only mounts `POST` and `DELETE`. Clients that perform the **GET handshake** first — notably **Claude Code**, but also Cursor and other streamable-HTTP clients — fail to connect: the server is silently dropped and none of the WhatsApp tools are exposed.

## Environment

- GOWA: `v9.3.1` (and identical code on `main`)
- `mcp-go` dependency: `mark3labs/mcp-go` `v0.56.0`
- Transport: streamable HTTP at `/mcp`, stateless, streaming disabled (`WithStateLess(true)` + `WithDisableStreaming(true)`), behind basic auth
- Code: `src/ui/mcp/route.go` — `Register()`

## Reproduction

Using any streamable-HTTP MCP client that does a GET-first handshake:

```
$ curl -i -X POST https://host/mcp -H 'Content-Type: application/json' \
    -H 'Accept: application/json, text/event-stream' \
    -H 'MCP-Protocol-Version: 2025-03-26' \
    -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-03-26","capabilities":{},"clientInfo":{"name":"test","version":"0.0.0"}}}'
HTTP/2 200
MCP-Protocol-Version: 2025-03-26
...
{"jsonrpc":"2.0","id":1,"result":{"protocolVersion":"2025-03-26","capabilities":{"tools":{},"resources":{}},"serverInfo":{"name":"WhatsApp Web Multidevice MCP Server","version":"v9.3.1"}}}

$ curl -i -X GET https://host/mcp -H 'Accept: application/json, text/event-stream' -H 'MCP-Protocol-Version: 2025-03-26'
HTTP/2 405 Method Not Allowed
allow: POST, DELETE
```

POST works, but the GET returns `405`. Claude Code connects to `https://mcp.sentry.dev/mcp` (a spec-compliant streamable-HTTP server) fine, while GOWA's endpoint is dropped at session load — so the symptom is "server configured but no tools appear."

## Root cause

In `src/ui/mcp/route.go`, `Get("/mcp")` is deliberately **not mounted** (see the comment in `Register()`). The streamable-HTTP transport (spec revision `2025-03-26` and later) requires the client to perform a `GET` first:

- The GET negotiates `MCP-Protocol-Version` and optionally establishes the stream used for server-initiated messages.
- Claude Code's client always performs this GET handshake before its first `POST initialize`. A non-2xx response aborts the connection.

Mounting only `POST`/`DELETE` is an implementation detail that leaks into the wire protocol and breaks real clients.

## Expected behavior

`GET /mcp` should return `200` with an `MCP-Protocol-Version` response header (matching `initialize`'s), so clients can complete the handshake and proceed to `POST`.

## Suggested fix

Mount `GET` alongside `POST`/`DELETE`, returning a minimal 200 response. Keep `WithDisableStreaming(true)` — the comment explains the fasthttp-adaptor goroutine/FD leak that streaming would introduce, so we don't want to re-enable it:

```go
router.Get("/mcp", func(c fiber.Ctx) error {
    c.Set("MCP-Protocol-Version", "2025-03-26")
    return c.SendString("")
})
router.Post("/mcp", handler)
router.Delete("/mcp", handler)
```

With `WithStateLess(true)` there are no server-initiated notifications, so an empty 200 GET response satisfies the spec.

Happy to open the PR if maintainers want it.

triasrahman:
Fixes #839

## Problem

Streamable-HTTP clients — notably **Claude Code** and Cursor — perform a `GET /mcp` before their first `POST initialize` to agree on `MCP-Protocol-Version`. GOWA only mounted `POST`/`DELETE`, so the GET returned `405` and the client dropped the server silently (no WhatsApp tools appear).

## Fix

Mount `GET /mcp` returning a minimal `200` with the `MCP-Protocol-Version` response header. This is stateless and streaming is already disabled, so there are no server-initiated notifications to stream — an empty 200 satisfies the spec for whitespace handshake consumers.

`WithDisableStreaming(true)` is kept as-is: re-enabling streaming would pin goroutines/FDs through the fasthttp adaptor (see the comment in `Register()`).

## Test

- `GET /mcp` now returns `200` (was `405`) with `MCP-Protocol-Version: 2025-03-26`.
- `POST initialize` unchanged (`200`, `serverInfo: WhatsApp Web Multidevice MCP Server`).

### DIFF
--- a/src/cmd/mcp_route_test.go
+++ b/src/cmd/mcp_route_test.go
@@ -98,8 +98,9 @@ func TestMcpEndpointRequiresBasicAuth(t *testing.T) {
 // TestMcpEndpointGetDoesNotHang guards C1: a GET on /mcp must not reach
 // mcp-go's standalone SSE notification stream, which blocks forever behind
 // the fasthttp adaptor (ctx.Done() only fires on server shutdown, never on
-// client disconnect). Only POST and DELETE are mounted, so GET must come
-// back as a client error without app.Test's default timeout ever tripping.
+// client disconnect). Streaming is disabled and the server is stateless, so
+// the GET handshake is answered immediately with a minimal 200; app.Test's
+// default timeout below guards that it does not hang.
 func TestMcpEndpointGetDoesNotHang(t *testing.T) {
 	app := newMcpTestApp(false)
 
@@ -109,8 +110,11 @@ func TestMcpEndpointGetDoesNotHang(t *testing.T) {
 	require.NoError(t, err)
 	defer resp.Body.Close()
 
-	assert.True(t, resp.StatusCode == fiber.StatusNotFound || resp.StatusCode == fiber.StatusMethodNotAllowed,
-		"GET /mcp returned %d, want 404 or 405", resp.StatusCode)
+	assert.Equal(t, fiber.StatusOK, resp.StatusCode, "GET /mcp returned %d, want 200", resp.StatusCode)
+	assert.Equal(t, "2025-03-26", resp.Header.Get("MCP-Protocol-Version"))
+	body, err := io.ReadAll(resp.Body)
+	require.NoError(t, err)
+	assert.Empty(t, body)
 }
 
 // TestMcpToolCallRejectsInvalidArgumentsViaSchema guards I1: with
--- a/src/ui/mcp/route.go
+++ b/src/ui/mcp/route.go
@@ -59,9 +59,15 @@ func Register(router fiber.Router, dm *whatsapp.DeviceManager, deps Deps) {
 
 	handler := adaptor.HTTPHandler(httpServer)
 	// POST carries JSON-RPC calls; DELETE is part of the streamable-HTTP
-	// session lifecycle. GET is intentionally not mounted: with streaming
-	// disabled mcp-go would just 405 it, so Fiber's own 404 for an
-	// unmounted method is equivalent and keeps the route surface narrow.
+	// session lifecycle. GET is required by the streamable-HTTP handshake:
+	// clients such as Claude Code or Cursor perform a GET first to agree on
+	// MCP-Protocol-Version before their initial POST, and drop the server on
+	// any non-2xx. With streaming disabled (and the server stateless) there
+	// are no server-initiated messages, so a minimal 200 is spec-compliant.
+	router.Get("/mcp", func(c fiber.Ctx) error {
+		c.Set("MCP-Protocol-Version", "2025-03-26")
+		return c.SendString("")
+	})
 	router.Post("/mcp", handler)
 	router.Delete("/mcp", handler)
 }


aldinokemal:
Thanks for digging into this, and sorry to close it — but I checked the change against the MCP spec and the official client SDK rather than just the test, and it turns out the pre-PR behaviour was already correct while this makes it non-compliant.

### GET is not a handshake, and an empty 200 is not a legal answer

MCP Streamable HTTP spec, *Listening for Messages from the Server*:

> 1. The client **MAY** issue an HTTP GET to the MCP endpoint. This can be used to open an SSE stream...
> 3. The server **MUST** either return `Content-Type: text/event-stream` in response to this HTTP GET, or else return HTTP 405 Method Not Allowed, indicating that the server does not offer an SSE stream at this endpoint.

GET is optional and exists to open an SSE stream. Item 3 is a hard MUST with exactly two legal outcomes, and this PR returns a third: **200 with `Content-Length: 0` and no `Content-Type` at all**.

`MCP-Protocol-Version` is also a *client request* header, not a server response header — it is negotiated in the `initialize` **result body** and then echoed by the client on subsequent POSTs. Nothing is agreed via GET.

### The library was already doing the right thing

`mark3labs/mcp-go@v0.56.0`, `server/streamable_http.go`:

```go
// WithDisableStreaming prevents the server from responding to GET requests with
// a streaming response. Instead, it will respond with a 405 Method Not Allowed status.
```

and `handleGet` at `:742` does exactly that. With only POST/DELETE mounted, Fiber returned `405 Method Not Allowed` with `Allow: POST, DELETE` — precisely what item 3 mandates. The 405 in #839's reproduction is the spec-correct response, not the bug.

### The change induces a reconnect loop in the official SDK

Traced through `modelcontextprotocol/typescript-sdk`, `client/streamableHttp.ts`:

1. `GET /mcp` → 200, so `response.ok` is true and the 405 branch at `:627-640` — which exists precisely to tolerate this — is skipped. `_handleSseStream(..., isReconnectable = true)` runs instead, with no `Content-Type` validation on the GET path.
2. The empty stream ends immediately → `canResume` and `needsReconnect` both true → `_scheduleReconnection(opts, 0)` with **`attemptCount` hardcoded to 0**.
3. That path *resolves* rather than throwing, so the `.catch()` that increments `attemptCount` never fires and `maxRetries: 2` is never reached. Delay is `initialReconnectionDelay` = 1000ms.

Net result: every connected spec-compliant client polls `GET /mcp` about once a second, forever. With the old 405 the SDK returns cleanly and never polls.

Separately, the hardcoded `2025-03-26` is the spec's *fallback for servers that receive no header*, not a version this server negotiates — mcp-go is on `LATEST_PROTOCOL_VERSION = "2025-11-25"`. Against the running server, `initialize` returned `2025-06-18` / `2025-11-25` / `2025-11-25` for three different client versions while the GET header claimed `2025-03-26` in all three.

### On #839 itself

The premise that clients GET before `POST initialize` does not hold — in the SDK the standalone GET only starts once the `initialized` notification is 202-accepted, i.e. strictly after initialize. The reproduction transcript in #839 also does not match what this server actually emits (it shows `"capabilities":{"tools":{},"resources":{}}` and an `MCP-Protocol-Version` response header on POST; the real server sends `{"tools":{"listChanged":true}}` and no such header). So the underlying "no tools appear in Claude Code" symptom is still undiagnosed, and this change would not have fixed it — I'm keeping #839 open.

Credit where due: the auth wiring here was fine. I checked both registration paths (`cmd/rest.go:163` behind basic auth, and `useMcpOAuthMiddleware`'s method-agnostic `router.Use("/mcp", auth)`) and GET is correctly protected — unauthenticated GET returns 401. Build, vet and the full test suite were also green.

If we ever do want an explicit GET mount, the one-line version is `router.Get("/mcp", handler)` so mcp-go emits its own spec-correct 405 — the hang the old comment warned about doesn't apply, since `WithDisableStreaming(true)` short-circuits before the SSE loop.

Closing this one. Please do keep the reports coming.

PR #841, opened 2026-09-13
feat(webhook): let a per-device webhook add to the global targets instead of replacing them

Kizuno18:
Adds an opt-in flag, `WHATSAPP_WEBHOOK_DEVICE_MERGE_GLOBAL` / `--webhook-device-merge-global`, so a device that has its own `webhook_url` still delivers its events to the global `WHATSAPP_WEBHOOK` URLs. Default stays `false`, so nothing changes for anyone who does not set it.

## Why

Today `PATCH /devices/{device_id}/webhook` is a replacement: once a device has a `webhook_url`, `forwardPayloadToConfiguredWebhooks` only ever hits that one URL. That is the right default for a tenant that wants isolation, but it bites when two services share an instance and one of them claims a device. In our stack an alarm service registers its own per-device webhook (`webhook_events: "message"`) on a handful of devices, and from that moment the two global consumers (an auto-responder and an events sink) silently stop receiving anything from those devices. Nothing in the logs says so either: the forwarder happily reports `Forwarding message to 1 configured webhook(s)`.

## What it does

With the flag on and a device override present, the forwarder delivers in two legs:

- the device URL keeps the device's `webhook_secret`, `webhook_events` and `webhook_insecure_skip_verify` (unchanged);
- the global URLs are signed with the global secret and filtered by `WHATSAPP_WEBHOOK_EVENTS`, i.e. exactly what they would get from a device without an override.

The ignore-JID rules (`WHATSAPP_WEBHOOK_IGNORE_JIDS` plus the per-device `webhook_ignore_groups` override) gate both legs, a global URL equal to the device URL is delivered once, and the error contract is the same as `forwardToWebhooks`: an error only comes back when every attempted target failed. Chatwoot forwarding is untouched. The session-id enrichment now also runs when only the global leg is allowed, since those consumers rely on it.

Implementation is contained in `webhook_forward.go` (`globalWebhookURLsExcluding` plus the second `forwardToWebhooks` call), one config field, the viper/cobra wiring in `cmd/root.go`, and docs (`readme.md`, `docs/webhook-payload.md`, `.env.example`).

## Tests

`webhook_forward_device_merge_test.go` drives `forwardPayloadToConfiguredWebhooks` through the existing `webhookStorageForTest` / `submitWebhookFn` seams and covers: flag off keeps device-only delivery, flag on delivers to device and global with the right secret on each, each leg honours its own event whitelist (including a device with no `webhook_events` inheriting the global list), shared URL dedup, and the all-legs-failed error rule. `go build`, `go vet` and `go test ./...` are green.


Kizuno18:
Closing: this branch was cut from our integration branch and drags in unrelated commits. Reopening from a clean branch on top of main.

PR #842, opened 2026-09-13
feat(webhook): let a per-device webhook add to the global targets instead of replacing them

Kizuno18:
Adds an opt-in flag, `WHATSAPP_WEBHOOK_DEVICE_MERGE_GLOBAL` / `--webhook-device-merge-global`, so a device that has its own `webhook_url` still delivers its events to the global `WHATSAPP_WEBHOOK` URLs. Default stays `false`, so nothing changes for anyone who does not set it. (Replaces #841, which was cut from the wrong base.)

## Why

Today `PATCH /devices/{device_id}/webhook` is a replacement: once a device has a `webhook_url`, `forwardPayloadToConfiguredWebhooks` only ever hits that one URL. That is the right default for a tenant that wants isolation, but it bites when two services share an instance and one of them claims a device. In our stack an alarm service registers its own per-device webhook (`webhook_events: "message"`) on a handful of devices, and from that moment the two global consumers (an auto-responder and an events sink) silently stop receiving anything from those devices. Nothing in the logs says so either: the forwarder happily reports `Forwarding message to 1 configured webhook(s)`.

## What it does

With the flag on and a device override present, the forwarder delivers in two legs:

- the device URL keeps the device's `webhook_secret`, `webhook_events` and `webhook_insecure_skip_verify` (unchanged);
- the global URLs are signed with the global secret and filtered by `WHATSAPP_WEBHOOK_EVENTS`, i.e. exactly what they would get from a device without an override.

`WHATSAPP_WEBHOOK_IGNORE_JIDS` gates both legs, a global URL equal to the device URL is delivered once, and the error contract is the same as `forwardToWebhooks`: an error only comes back when every attempted target failed. Chatwoot forwarding is untouched. The session-id enrichment now also runs when only the global leg is allowed, since those consumers rely on it.

Implementation is contained in `webhook_forward.go` (`globalWebhookURLsExcluding` plus the second `forwardToWebhooks` call), one config field, the viper/cobra wiring in `cmd/root.go`, and docs (`readme.md`, `docs/webhook-payload.md`, `.env.example`).

## Tests

`webhook_forward_device_merge_test.go` drives `forwardPayloadToConfiguredWebhooks` through the existing `webhookStorageForTest` / `submitWebhookFn` seams and covers: flag off keeps device-only delivery, flag on delivers to device and global with the right secret on each, each leg honours its own event whitelist (including a device with no `webhook_events` inheriting the global list), shared URL dedup, and the all-legs-failed error rule. `go build`, `go vet` and `go test ./infrastructure/whatsapp ./cmd` are green.


src/infrastructure/whatsapp/webhook_forward_device_merge_test.go
```diff
@@ -0,0 +1,144 @@
...
+func runDeviceMergeForward(t *testing.T, merge bool, globalURLs []string, globalEvents []string, deviceURL, deviceEvents, eventName string, failURLs map[string]bool) ([]mergeDelivery, error) {
+	t.Helper()
+
+	originalWebhooks := config.WhatsappWebhook
+	originalEvents := config.WhatsappWebhookEvents
+	originalIgnore := config.WhatsappWebhookIgnoreJids
+	originalMerge := config.WhatsappWebhookDeviceMergeGlobal
+	config.WhatsappWebhook = globalURLs
+	config.WhatsappWebhookEvents = globalEvents
+	config.WhatsappWebhookIgnoreJids = nil
```
aldinokemal, line 30: Could we add a test covering `WHATSAPP_WEBHOOK_IGNORE_JIDS` with merge mode enabled?

`jidAllowed` now gates both the device and global legs, but the new merge tests always reset `WhatsappWebhookIgnoreJids` to `nil`, so this part of the new routing behavior is not exercised.

A small case where the payload matches an ignored JID while both device and global URLs are configured would lock in the intended contract that the ignore rule suppresses both legs.

aldinokemal (commented):
One small test coverage suggestion.

aldinokemal (changes requested):
Good feature, and the flag-off safety property holds — I checked that first, since it is what decides whether this is risky for existing deployments.

**Flag off is provably unchanged.** With `WhatsappWebhookDeviceMergeGlobal == false`, `globalURLs` stays nil so `globalAllowed` is false: the reordered `&&` operands are both pure so the boolean is identical, the skip guard reduces to the old one, `if webhookAllowed || globalAllowed` reduces to `if webhookAllowed`, and the whole merge block is dead. Default confirmed at runtime (cobra `DefValue="false"`, `settings.go:53`). Env/viper/cobra naming is consistent across `readme.md`, `docs/webhook-payload.md`, `.env.example` and `root.go:187,462` — verified live, `IsSet=true GetBool=true`, and it correctly uses the `IsSet`+`GetBool` pattern rather than the old workaround.

Clean everywhere: `go build`, `go vet`, `gofmt -l`, `go test ./...` (13 pkgs ok), `go test -race ./infrastructure/whatsapp ./cmd` ok. Merges cleanly. The new tests drive the real `forwardPayloadToConfiguredWebhooks` through the genuine `webhookStorageForTest` / `submitWebhookFn` seams — not tautological.

**Ignore-JID: the contract you asked about holds.** I wrote the test — `WHATSAPP_WEBHOOK_IGNORE_JIDS=@g.us`, merge on, both a device and a global URL, group payload → **0 deliveries on both legs**. So it is a coverage gap, not a bug. Still worth committing so it stays locked in.

**The failing Kilo check is a flake, not a finding:** `"Review failed: The model output limit was reached"`, 2h20m, `annotations_count: 0`, `output.text: null`. Ignore it.

Two defects, both opt-in-only, so neither threatens anyone on defaults.

### D1 — the device leg starves the global leg on the shared context

`webhook_forward.go:143-162` runs both legs sequentially on the same `ctx`. Every caller allows 30s, and `submitWebhook` (`webhook.go:67-96`) does 5 attempts with **non-ctx-aware** `time.Sleep` backoff of 1+2+4+8 = 15s plus a 10s per-request timeout. A slow or hung device webhook therefore consumes the entire budget and the global leg runs on an already-dead context.

Measured with a real `httptest` pair and the real `submitWebhook`:

```
elapsed=33.011s
err=all 1 webhook(s) failed for message; global: all 1 webhook(s) failed for message
globalHits=0
```

A perfectly healthy global endpoint got nothing. In production ~3 failed 10s attempts on the device URL guarantees it. This is the scenario the feature exists for — "keep the global hub fed even when the device has its own endpoint" — so I'd rather have it fixed before merge than after. Running the global leg concurrently, or deriving a per-leg deadline, should do it.

### D2 — unconditional dedup drops a shared URL when the device filter rejects the event

`webhook_forward.go:135` computes `globalURLs = globalWebhookURLsExcluding(webhookURLs)` regardless of `webhookAllowed`, so a global URL that string-equals the device URL is excluded even when the device leg will never deliver. With `WHATSAPP_WEBHOOK=https://hub.example/in` (no event filter), device `webhook_url` the same and device `webhook_events: message`, event `message.ack`:

```
deliveries=[] (len=0)
```

The global config subscribes to that URL for all events and receives nothing — which contradicts the line in `docs/webhook-payload.md` promising the global URLs keep `WHATSAPP_WEBHOOK_EVENTS`. Only exclude the device URLs when `webhookAllowed` is true.

### Nits

- Dedup is exact-string (`slices.Contains`). Global `https://hub.example/in/` + device `https://hub.example/in` → **2 deliveries to the same endpoint**, one per secret. The docs do say "equals", so it is documented-as-is, but a trailing slash or host-case difference silently double-delivers.
- `webhook_forward.go:147` still logs `"Skipping event %s for configured webhooks, but allowing Chatwoot"` when the global leg is about to run and Chatwoot may be disabled.
- `"Forwarding %s to %d configured webhook(s)"` now fires twice per event in merge mode — noisy at INFO.
- Docs say the device leg is "signed with the device's `webhook_secret`"; `webhook.go:28` falls back to the global secret when it is empty. Worth stating.

**Error contract verified correct** in every combination, including the case I most expected to be wrong: a zero-target leg is never counted as failed (`forwardToWebhooks` returns nil on `total == 0`), and there is no data race — `addWebhookSessionID` mutates once before either leg, and both legs are fully synchronous.

D1 is the one blocking this for me; D2 and the ignore-JID test are small. Happy to merge once D1 is addressed.

PR #843, opened 2026-09-14
fix(device): make DELETE purge reliable and decode path device_id

hanlest:
## Summary

Fixes two device-management bugs seen with gowa-ui and Fiber v3:

1. **DELETE /devices/{id} could appear to succeed while the slot stayed listed** — `PurgeDevice` called `cli.Logout()`, which fires `LoggedOut` and the keep-slot callback (`keepSlotLogout`) introduced in #728. That contradicted full purge semantics.
2. **DELETE returned 404 for device IDs containing spaces** — gowa-ui calls `encodeURIComponent(device.id)` (e.g. `Estar%20Siempre`), but Fiber v3 leaves percent-escapes in `c.Params("device_id")`, so the registry lookup used the wrong key.

## Changes

- `PurgeDevice`: resolve slot via `ResolveDevice`, clear `OnLoggedOut` before logout, delete chatstorage for both slot id and JID partition keys when they differ, use resolved id for registry removal.
- `RemoveDevice` usecase: map missing device to `404`, broadcast `DEVICE_REMOVED` with updated device list (mirrors logout websocket refresh).
- REST: `pathDeviceID()` (`url.PathUnescape`) for all `:device_id` routes (devices + chatwoot config).

## Test plan

- [x] `go test -tags purego ./infrastructure/whatsapp/ -run PurgeDevice`
- [x] `go test -tags purego ./ui/rest/ -run PathDeviceID`
- [x] Manual: create device id `Estar Siempre`, DELETE from dashboard, confirm it disappears from `GET /devices`

## Related

- Context: #728 (logout keeps slot vs DELETE purges)


src/infrastructure/whatsapp/device_manager.go
```diff
@@ -266,6 +266,11 @@ func (m *DeviceManager) PurgeDevice(ctx context.Context, deviceID string) error
 		return fmt.Errorf("device id is required")
 	}
 
+	inst, resolvedID, err := m.ResolveDevice(deviceID)
+	if err != nil {
```
aldinokemal, line 325: **P1 — preserve exact slot identity for destructive operations.** `ResolveDevice` trims `deviceID` before lookup, while `CreateDevice` stores the requested ID verbatim. That means two valid slots such as `"sales"` and `" sales "` can coexist, but deleting the latter will resolve and purge `"sales"` instead. The REST helper now correctly decodes `%20`, so this can be reached through the API. Please prefer an exact ID lookup first (or normalize IDs consistently at creation time with an explicit migration/backward-compatibility story) before falling back to JID resolution.

src/infrastructure/whatsapp/device_manager.go
```diff
@@ -275,46 +280,54 @@ func (m *DeviceManager) PurgeDevice(ctx context.Context, deviceID string) error
...
 	// Delete chatstorage data for this device (local cleanup — surfaced on failure).
+	// Chats/messages are partitioned by the WhatsApp JID once paired, while the
+	// registry slot keeps its own id — purge both keys when they differ.
 	if m.storage != nil {
-		if err := m.storage.DeleteDeviceData(deviceID); err != nil {
-			logrus.WithError(err).Warnf("[DEVICE_MANAGER] failed to delete chatstorage for device %s", deviceID)
-			recordErr(err)
+		partitionKeys := []string{resolvedID}
+		if jidKey := strings.TrimSpace(inst.JID()); jidKey != "" && jidKey != resolvedID {
+			partitionKeys = append(partitionKeys, jidKey)
```
aldinokemal: **P1 — this can delete data belonging to another companion slot.** `inst.JID()` is the non-AD/bare-number JID, and multiple companion sessions for the same number may share it while having different AD JIDs. Chat storage is written with that non-AD JID as `device_id`, and `DeleteDeviceData(jidKey)` removes every row for that partition. Purging one slot can therefore wipe chats/messages for a sibling slot that still uses the same number. Please only delete this JID partition when no other registered slot claims it (or move chat-storage partitioning to a per-slot identity). Add a regression test with two slots sharing the same non-AD JID and verify the sibling data survives.

src/infrastructure/whatsapp/device_manager.go
```diff
@@ -324,7 +337,7 @@ func (m *DeviceManager) PurgeDevice(ctx context.Context, deviceID string) error
 	recordErr(m.deleteStoreRowsForJID(ctx, jid))
 
 	// Remove from registry last
-	m.RemoveDevice(deviceID)
+	m.RemoveDevice(resolvedID)
```
aldinokemal: **P2 — a transient cleanup failure becomes non-retryable.** `firstErr` can contain a local cleanup failure, but the slot is still removed here before that error is returned. On the next DELETE, the new `ResolveDevice` requirement fails immediately because the registry entry is gone, so the failed cleanup cannot be retried. Previously the purge path could still attempt cleanup by the supplied ID without a live slot. Please keep enough persisted identity/state to retry cleanup, or only remove the registry entry after all required local cleanup succeeds (while keeping the operation idempotent). A fail-once/succeed-on-retry test would catch this.

aldinokemal (changes requested):
Thanks for tackling the delete/purge and encoded-device-id issues. I found three correctness regressions in the current implementation that should be addressed before merge:

1. Purging by the shared non-AD JID can delete chat history owned by a sibling companion slot using the same phone number.
2. Routing DELETE through `ResolveDevice` changes destructive lookup semantics because `ResolveDevice` trims the ID while `CreateDevice` preserves the requested ID verbatim; an encoded ID with leading/trailing spaces can therefore resolve to a different slot.
3. Local cleanup failures remove the registry entry before returning the error, and the new required `ResolveDevice` step prevents a subsequent DELETE from retrying that cleanup.

I left inline comments on each path with concrete failure scenarios and suggested fixes. I would also add regression coverage for two companion slots sharing one non-AD JID, exact-vs-trimmed device IDs, and retry after a transient cleanup failure.

src/infrastructure/whatsapp/device_manager.go
hanlest: Addressed in d379aae: PurgeDevice now uses resolveDeviceForPurge (exact registry lookup, no trim) before JID fallback. Test: TestPurgeDevice_ExactSlotIDDoesNotTrimToSibling.

src/infrastructure/whatsapp/device_manager.go
hanlest: Addressed in d379aae: the non-AD JID partition is deleted only when nonADJIDClaimedByOtherSlot finds no other slot (registry + ListDeviceRecords) with the same bare JID. Test: TestPurgeDevice_DoesNotDeleteSharedJIDPartitionWhenSiblingSlotExists.

hanlest:
@aldinokemal Thanks for the detailed review — pushed d379aae addressing all three points (exact purge ID lookup, shared non-AD JID partition guard, retryable local cleanup) with regression tests. Ready for another look when you have a moment.

src/infrastructure/whatsapp/device_manager.go
hanlest: Addressed in d379aae: RemoveDevice runs only after local cleanup succeeds (no recorded errors), so DELETE can be retried. Tests: TestPurgeDevice_SurfacesLocalCleanupFailure and TestPurgeDevice_RetryAfterTransientCleanupFailure.

aldinokemal (changes requested):
Thanks for the fast turnaround on `d379aae` — all three of my earlier points are genuinely addressed, and I verified the Fiber premise rather than taking it on trust.

**Fiber v3 claim: confirmed correct.** `UnescapePath` defaults to `false` (`app.go:214`), `configDependentPaths()` only unquotes when it is set (`ctx.go:652-656`), and this repo never sets it. So `url.PathUnescape` is a *single* decode, not a double one. Verified through a real router:

```
/devices/Estar%20Siempre     -> "Estar Siempre"
/devices/Estar%2520Siempre   -> "Estar%20Siempre"   (correct round-trip for a literal %)
/devices/a+b                 -> "a+b"               (PathUnescape, not QueryUnescape — right call)
```

Build/vet/test are clean on `d379aae` (`go test ./...` all 13 packages ok; the 8 `PurgeDevice` tests and `TestPathDeviceID_DecodesPercentEncodedSpace` all pass), and it merges cleanly.

Holding at request-changes for two defects introduced by the hardening itself. Both are small, and the existing suite is blind to both because nothing asserts on which partitions get deleted.

### 1. `excludeSlotID` is trimmed, so a padded slot cannot exclude itself

`device_manager.go:213` does `excludeSlotID = strings.TrimSpace(excludeSlotID)`, but line 224 compares it to the untrimmed `rec.DeviceID` and line 235 to the untrimmed registry key `id`.

Purging the **sole** slot `" sales "`: `excludeSlotID` becomes `"sales"`, which matches neither, so the slot's own record reads as *another* claimant and the guard returns `true`:

```
warning msg="skipping shared chatstorage partition 6281700000001@s.whatsapp.net for slot  sales
             (another slot still claims this number)"
DeleteDeviceData called with partitions: [" sales "]
```

Control with a plain id deletes both partitions, which isolates the trim as the cause. Net effect: DELETE reports success, the slot disappears, and every chat/message for that number stays on disk permanently with no way to retry. That is the exact id shape this PR exists to support — `TestPurgeDevice_ExactSlotIDDoesNotTrimToSibling` uses `" sales "` but never asserts on partitions, so it sails past.

Fix: drop line 213.

### 2. A `ListDeviceRecords` failure is swallowed — this re-opens my P2

`device_manager.go:217-222` logs and returns `true` on error, so the caller cannot tell "claimed" from "lookup failed" and the error never reaches `firstErr`:

```
PurgeDevice err=<nil> | partitions deleted=["solo"] | slot still registered=false
```

Any transient DB error on that one call gives a 200, removes the slot, leaves the chat data, and makes retry impossible — `resolveDeviceForPurge` now hard-fails on a missing slot. That is the failure mode I objected to in P2, reached through the code added for P1.

Fix: return `(bool, error)` and feed it through `recordErr` so the slot survives for retry.

Both fixes want a regression test that asserts on the deleted partitions, which nothing currently does.

### Lower priority, not blocking

- `resolveConfigDeviceID` (`chatwoot_config.go:204`) and the non-destructive device handlers still go through the trimming `ResolveDevice`. Before this PR a request for `" sales "` 404'd; now that `%20` decodes, it silently resolves to `"sales"`. Same class as my P1, different path.
- `strings.Contains(err.Error(), "not found")` in `usecase/device.go:100-102` is fragile — correct today only because `resolveDeviceForPurge` is the sole producer of that text. A typed sentinel would be sturdier.
- `ValidateDeviceID` only rejects blank-after-trim, so the accepted charset just widened to include `/`, `..` and NUL. Not exploitable — I grepped for `filepath.Join|os.Open|os.Create|os.MkdirAll|os.Remove` intersected with device identifiers and got zero hits; `device_id` is only a map key and a SQL bind parameter. Worth a charset check as a follow-up.
- `DEVICE_REMOVED` broadcast matches the `DEVICE_LOGGED_OUT` contract exactly. It emits the *requested* id rather than `resolvedID`, so a purge-by-JID broadcasts a JID as `device_id` — the `devices` array is authoritative so UIs recover, but worth aligning.

The rest — the exact-lookup fix, the shared-JID guard's intent, the retry ordering, the broadcast, the decode direction — all check out. Two small commits and this is good to go.

PR #845, opened 2026-09-16
fix(history-sync): name 1:1 chats from the contact store, not DisplayName

### LINKED ISSUE #844: History sync stores phone numbers and the owner's own push name as 1:1 chat names
`processConversationMessages` names each synced chat from `conv.GetDisplayName()` and never consults the contact store, so two things end up in `chats.name`:

- the bare user part of the JID, when `DisplayName` is empty — even for contacts that are in the address book;
- **the account owner's own push name**, for conversations where WhatsApp puts it in `DisplayName`, which then becomes the stored name of somebody else's chat.

Measured on one real account: **359 of 809 one-to-one chats (44%) carry a name that does not match the contact**, 165 of them one of the two cases above. Several unrelated chats were all stored under the owner's own push name.

### Why it is not only cosmetic

`usecase/chat.go` already repairs the name at read time through `ChatDisplayNameResolver`, so the chat *displays* correctly. But **chat search filters on the stored name in SQL, before the resolver runs**. A chat stored under a phone number or the owner's name can never be found by searching the contact's name.

Through the MCP server that makes the contact effectively unreachable: search is how you obtain the JID, and the JID is what you need to read the conversation. Searching a contact by name returns nothing even though the chat exists and has thousands of messages in it.

The read-time resolver also cannot repair the owner's-name case at all. `isJIDFallbackName` only treats empty or JID-like names as replaceable, and a real name is neither, so those rows stay wrong permanently — a re-sync does not fix them either.

### Suggested fix

The machinery already exists: have the history sync ask the same `ChatDisplayNameResolver` the read path uses, before falling back to `DisplayName`. PR incoming.

arivelli:
Fixes https://github.com/aldinokemal/go-whatsapp-web-multidevice/issues/844

A conversation's `DisplayName` cannot be trusted as a one-to-one chat name. It arrives empty for many conversations, and for some it carries the account owner's own push name, which then gets stored as the name of somebody else's chat.

Measured on one real account: 359 of 809 one-to-one chats ended up with a name that does not match the contact, and several unrelated chats were all stored under the owner's own push name.

Neither case is repairable afterwards:

- `usecase/chat.go` already resolves names at read time through `ChatDisplayNameResolver`, but `isJIDFallbackName` only replaces names that look like a JID or a bare number — a real name like the owner's sticks permanently.
- Chat search filters on the stored name in SQL, before any resolver runs, so those chats cannot be found by the contact's name at all. Through the MCP server that makes the contact unreachable, since search is how you obtain the JID.

### The change

Ask the same resolver the read path already uses, before falling back to `DisplayName`, passing no stored name so it cannot short-circuit on one. When the contact store knows nobody it answers the bare user part, which is no better than `DisplayName`, so `DisplayName` is kept in that case.

The naming decision is extracted into `conversationChatName` so it can be tested directly; the resolver is built once per sync rather than per conversation, since it caches contacts.

### Tests

Three cases in `history_sync_name_test.go`: the contact wins over an owner's push name arriving as `DisplayName`, an empty `DisplayName` falls back to the contact, and an unknown contact keeps the incoming `DisplayName`. Plus a nil-resolver no-op, since `processConversationMessages` is called with a nil client in the existing tests.

`go build ./...` and `go test ./infrastructure/whatsapp/...` pass.


aldinokemal (approved):
Reviewed against `4338e31`. Verified locally, all from `src/`:

```
go build ./...                              exit 0
go vet ./...                                exit 0
go test ./...                               13 packages ok, 0 failures
go test -race ./infrastructure/whatsapp/... ok (3.240s)
```

Merges cleanly onto `main`. The three new tests are real, not tautological — they genuinely pin the contact-beats-DisplayName case, the empty-DisplayName fallback and the nil-resolver guard.

### On CodeRabbit's group/newsletter comment

The mechanism is real, the severity isn't. `conversationChatName` passes `storedName: ""`, so `Resolve` returns `"Group <id>"` / `"Newsletter <id>"` for those servers — I reproduced that directly:

```
GROUP      -> "Group 120363012345678901"      (DisplayName was "Family Group")
NEWSLETTER -> "Newsletter 120363111111111111" (DisplayName was "Tech Channel")
```

But the only consumer, `GetChatNameWithPushNameByDevice`, hard-codes those same two strings and ignores the `pushName` argument entirely for `g.us` / `newsletter` (`sqlite_repository.go:1854,1858`). So the real group subject was already being discarded before this PR, and there is no reachable regression. The one state that differs needs a group chat row already named with its bare group id, and nothing in this repo ever writes one.

So: latent trap rather than a functional bug — but a genuine one, since `sqlite_repository.go:1810` explicitly invites someone to make that branch honour group info later. I'll land the guard as a follow-up on `main`:

```go
if jid.Server == types.GroupServer || jid.Server == types.NewsletterServer {
    return displayName
}
```

### Two notes for the record

- **This fix is forward-looking, not retroactive.** `GetChatNameWithPushNameByDevice` only replaces a stored name that looks like a bare JID user part, so the 359 already-corrupted rows measured in #844 stay corrupted on an existing install. Verified: a 1:1 chat stored as the owner's push name is *not* repaired by a re-sync; one stored as a bare number is. Users hitting #844 will still need to rebuild `chats`.
- **Minor:** `history_sync_name_test.go:10` is gofmt-dirty (the one-line struct wants expanding). Harmless, and I'll sweep it with the guard above.

Approving — thanks for the careful writeup and the measurements in the issue, they made this easy to verify.

